PR #2, opened 2016-11-02
add default https to link

pugovok:


faustbrian:
Do you have a link to docs where this is described?

pugovok:
https://vk.com/dev/api_requests

https – 1 – allows to get https links for photos and other media. 0 – methods return http links (default). Keep in mind that this parameter works only if you send requests to API through https.

PR #4, opened 2016-11-09
Fix the Uber provider

brunoti:
- The real issue was not solved. The scope should be sent as a string
but with "spaces" as separator.

Sorry about that, I thought that the "array way" was right, but then i've tried to add more scopes and I see that I was wrong.


faustbrian:
Thanks, please start to use more descriptive PR titles since `fix the uber provider` is pretty meaningless.

brunoti:
Ok!

Issue #13, opened 2016-12-07
Microsoft Live / Azure issue


cristianofromagio:
When using the current version of Microsoft Live provider, I was facing this problem:

`Client error: 'POST https://login.live.com/oauth20_token.srf' resulted in a `400 Bad Request` response:
{"error":"invalid_request","error_description":"The provided request must include a 'code' input parameter."}
in RequestException.php line 111`

So I went out to figure out what was causing this issue. I checked the redirects URIs on the Microsoft Application Registration Portal and they were the same. As the returned url was pointing out [`http://localhost:8000/auth/live/callback?error=unauthorized_client&error_description=The client does not have a secret configured. If you are the client application developer, configure a secret through the application management site at https://apps.dev.microsoft.com/.&state=8gIw(...)`], I was using the PRIVATE KEY instead of the PASSWORD [PASSWORD should be the correct one] as the LIVE_SECRET on .env, but fixing it didn't solve the issue either.

Digging deeper, found someone facing [similar](https://social.msdn.microsoft.com/Forums/onedrive/en-US/4f6aaeda-2bac-4d1c-8898-e99725526fdb/getting-status-code-400-bad-request-error-from-oauth20tokensrf-during-authorization-code-grant?forum=messengerconnect) issue, tried to learn more about the API flow to understand it's functions I found that they were replaced by a newer one.

They [converged](https://graph.microsoft.io/en-us/docs/authorization/converged_auth) the multiple authorization (authentication consequently) as one service using the Microsoft Graph and the Azure AD v2.0 endpoint. So, from there, I was able to dig into the messy Microsoft documentation and find the solution. Using the v2.0 endpoints for Azure AD listed [here](https://dev.office.com/hands-on-labs/4585) and [here](https://github.com/microsoftgraph/php-connect-rest-sample/blob/master/src/Constants.php) and being guided by [this tutorial](http://simonjaeger.com/microsoft-graph-authentication-with-the-converged-model-preview/).

Just as documentation:
[Authorize access to web applications using OAuth 2.0 and Azure Active Directory](https://docs.microsoft.com/en-us/azure/active-directory/active-directory-protocols-oauth-code)
[OAuth 2.0 Authorization Code Flow](https://docs.microsoft.com/en-us/azure/active-directory/active-directory-v2-protocols-oauth-code)
[...]

PR #14, opened 2016-12-07
Fixes SocialiteProviders/Providers#13

cristianofromagio:


faustbrian:
Tagged `v3.0.0` since this is a breaking change.

PR #17, opened 2016-12-19
Accliam provider is added

pavankumarkatakam:
Acclaim is a Badge management system for certified courses.


faustbrian:
This Provider also has to be added to https://github.com/SocialiteProviders/Providers/blob/master/build/split.sh before I merge it

pavankumarkatakam:
Done.

PR #19, opened 2016-12-27
Compatibility for signed Instagram API requests

eh-steve:


faustbrian:
Why would this be required? Some infos would help why this PR was send.

eh-steve:
If "Enforce signed requests" is enabled on the Instagram API, the OAuth flow will fail because the "sig" param is missing. Including the sig param even if "Enforce signed requests" is disabled doesn't affect anything, so it's best to include it to allow compatibility with signed requests.

See https://www.instagram.com/developer/secure-api-requests/

PR #23, opened 2017-01-17
Check if no dropbox avatar exists

iemand002:
Forgot to check what it does if there is no avatar => $user['profile_photo_url'] does not exist so check if isset


faustbrian:
Please use array_get for this check.

iemand002:
Never used that before, learned something new today

faustbrian:
You can omit the `, null` at the end, array_get returns null by default.

iemand002:
Finally 😄

PR #25, opened 2017-01-25
Fixes a type error in Jira integration

slava-vishnyakov:
Without this fix, there is a type error in signature of baseString function:

```
Type error: Argument 1 passed to SocialiteProviders\Jira\RsaSha1Signature::baseString() 
must be an instance of SocialiteProviders\Jira\Url, 
instance of GuzzleHttp\Psr7\Uri given, 
called in src/Jira/RsaSha1Signature.php on line 26
```

### DIFF
--- a/src/Jira/RsaSha1Signature.php
+++ b/src/Jira/RsaSha1Signature.php
@@ -2,6 +2,7 @@
 
 namespace SocialiteProviders\Jira;
 
+use GuzzleHttp\Psr7\Uri;
 use League\OAuth1\Client\Signature\Signature;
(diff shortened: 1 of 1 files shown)


slava-vishnyakov:
Previously to work around this problem, I've had to use someone's fork:

https://github.com/oschettler/Jira/blob/patch-1/src/RsaSha1Signature.php#L82

slava-vishnyakov:
@faustbrian Awesome, thanks! Could you please bump https://github.com/SocialiteProviders/Jira/ to `2.1.2` or something like that?

slava-vishnyakov:
👍  Awesome, thank you for all your efforts!

PR #28, opened 2017-02-02
Shopify Provider

CullenIO:
Adds a Shopify Provider.

PR #30, opened 2017-02-03
Shopify Provider

CullenIO:
Fixed #28.


CullenIO:
Will come back to later..

faustbrian:
Those are way to many scopes you are requesting. The only scopes provider should request by default are ones to read the users profile.

PR #40, opened 2017-02-20
Throw guzzle client exception when slack api request fails

jshayes:
Based on the [Slack API documentation](https://api.slack.com/web), when an API request fails the global boolean property `ok` is set to false. The request responds with a 200 OK response. As a result, an unsuccessful API request does not cause any exception to be thrown. This causes an undefined index error to occur when trying to fetch the user.

For example, doing `Socialite::driver('slack')->stateless()->user();` without a code, or with an invalid code, will cause the following error

```
ErrorException in Provider.php line 64:
Undefined index: user
```

However, when doing the same thing with the google driver, `Socialite::driver('google')->stateless()->user();` will throw the following exception.

```
ClientException in RequestException.php line 111:
Client error: `POST https://accounts.google.com/o/oauth2/token` resulted in a `400 Bad Request` response:
{
"error" : "invalid_request",
"error_description" : "Missing required parameter: code"
}
```

This PR makes the slack provider throw guzzle client exception when the requests fail. On this branch, doing `Socialite::driver('slack')->stateless()->user();` with an invalid code with throw the following exception.
```
RequestException in RequestException.php line 111:
Unsuccessful request: `POST https://slack.com/api/oauth.access` resulted in a `200 OK` response:
{"ok":false,"error":"invalid_code"}
```

### DIFF
--- a/src/Slack/Provider.php
+++ b/src/Slack/Provider.php
@@ -2,6 +2,10 @@
 
 namespace SocialiteProviders\Slack;
 
+use GuzzleHttp\Client;
+use GuzzleHttp\HandlerStack;
+use Psr\Http\Message\ResponseInterface;
+use GuzzleHttp\Exception\RequestException;
 use SocialiteProviders\Manager\OAuth2\User;
 use Laravel\Socialite\Two\ProviderInterface;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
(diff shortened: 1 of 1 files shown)


src/Slack/Provider.php
```diff
@@ -26,6 +30,50 @@ class Provider extends AbstractProvider implements ProviderInterface
     protected $scopeSeparator = ',';
 
     /**
+     * Middleware that throws exceptions for non successful slack api calls
+     * "http_error" request option is set to true.
+     *
+     * @return callable Returns a function that accepts the next handler.
+     */
+    private function getSlackApiErrorMiddleware()
```
jshayes: For this I just followed this same structure as https://github.com/guzzle/guzzle/blob/master/src/Middleware.php#L52

PR #45, opened 2017-03-02
Intercom Provider

bryceadams:


faustbrian:
The Provider also needs to be added to https://github.com/SocialiteProviders/Providers/blob/master/build/split.sh

src/Intercom/Provider.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    const IDENTIFIER = 'INTERCOM';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = [''];
```
faustbrian: The `$scopes` array can be removed as it is just an empty array like already defined here https://github.com/laravel/socialite/blob/3.0/src/Two/AbstractProvider.php#L62.

bryceadams:
👍

faustbrian:
Thanks! Will be added to packagist in a few minutes.

bryceadams:
Thanks for the fast turnaround!

faustbrian:
https://packagist.org/packages/socialiteproviders/intercom

PR #47, opened 2017-03-07
Added Bitbucket provider

barooney:


faustbrian:
Doesn't Socialite itself have a Bitbucket provider?

PR #48, opened 2017-03-15
Add SciStarter Provider

ryanvade:
[SciStarter](https://scistarter.com) provides centralized access to Citizen Science projects.  This provider can use the OAuth2 login system provided by SciStater.  I ran php-cs-fixer before committing.


faustbrian:
Also needs to be added to https://github.com/SocialiteProviders/Providers/blob/master/build/split.sh

ryanvade:
@faustbrian Do I need to create SocialiteProviders/SciStarter ?

faustbrian:
No, I need to do that.

PR #53, opened 2017-04-03
Remove Slack's default scopes to avoid conflicts

nesk:
The default scopes for the Slack provider are an issue since some combinations are forbidden. Here is what I get when I try to redirect a user with the following code:

```php
Socialite::with('slack')->scopes(['channels:history', 'channels:read', 'users:read', 'emoji:read', 'team:read'])->redirect();
```

image

The removal of the default scopes is **a breaking changement!** Merge this carefully.

### DIFF
--- a/src/Slack/Provider.php
+++ b/src/Slack/Provider.php
@@ -18,11 +18,6 @@ class Provider extends AbstractProvider implements ProviderInterface
     const IDENTIFIER = 'SLACK';
 
     /**
-     * {@inheritdoc}
-     */
-    protected $scopes = ['identity.basic', 'identity.email', 'identity.team', 'identity.avatar'];
-
-    /**
      * The separating character for the requested scopes.
      *
      * @var string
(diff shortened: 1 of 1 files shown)


faustbrian:
Will the authentication process still work properly without those and without the user having to manually define scopes?

nesk:
No, but keeping those default scopes doesn't allow the user to use other scopes, which is an issue for advanced applications. We should prefer a small configuration required by the user since it allows to use other scopes and not being limited to the `identity.*ˋ ones.

faustbrian:
There should be a way for the default scopes to be set if the user doesn't define some himself as we will get flooded with issues otherwise because bunch of applications will just stop working if they were just using the basic scopes.

nesk:
Alright, I perfectly understand. I will push a new commit today (CET timezone) to handle this behaviour.

nesk:
I've overridden the `getScopes` method to return the original default scopes if the user didn't define some.

PR #56, opened 2017-04-05
[Slack] Handle properly "missing_scope" errors

nesk:
Getting user informations requires the `identity.*` scopes, however we might want to not add them to the scope list for various reasons.

Instead of throwing an exception on this error, we return an empty user.

This is a non-breaking changement.

### DIFF
--- a/src/Slack/Provider.php
+++ b/src/Slack/Provider.php
@@ -106,9 +106,24 @@ class Provider extends AbstractProvider implements ProviderInterface
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get(
-            'https://slack.com/api/users.identity?token='.$token
-        );
+        try {
+            $response = $this->getHttpClient()->get(
+                'https://slack.com/api/users.identity?token='.$token
+            );
+        } catch (RequestException $exception) {
+            // Getting user informations requires the "identity.*" scopes, however we might want to not add them to the
+            // scope list for various reasons. Instead of throwing an exception on this error, we return an empty user.
+
+            if ($exception->hasResponse()) {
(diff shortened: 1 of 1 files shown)


faustbrian:
Please use the `array_get` method instead of `Arr::get` to keep it consistent with all other providers.

nesk:
Done, I've rebased the commit.

src/Slack/Provider.php
```diff
@@ -106,9 +106,24 @@ protected function getTokenUrl()
...
+                'https://slack.com/api/users.identity?token='.$token
+            );
+        } catch (RequestException $exception) {
+            // Getting user informations requires the "identity.*" scopes, however we might want to not add them to the
+            // scope list for various reasons. Instead of throwing an exception on this error, we return an empty user.
+
+            if ($exception->hasResponse()) {
+                $data = json_decode($exception->getResponse()->getBody(), true);
+
+                if (Arr::get($data, 'error') === 'missing_scope') {
```
sevastyanovio:
```
FatalThrowableError in Provider.php line 120:
Class 'SocialiteProviders\Slack\Arr' not found
```

sevastyanovio: should be `if (\Arr::get($data, 'error') === 'missing_scope') {`

sevastyanovio:
```
FatalThrowableError in Provider.php line 120:
Class 'SocialiteProviders\Slack\Arr' not found
```

nesk:
Oops, I missed that one. I've rebased the commit once again to get rid of all the calls to `Arr::get`. This time it should be alright.

PR #58, opened 2017-04-05
clientId instead of client_id

ryanvade:


ryanvade:
@faustbrian my apologies for the 'last minute' pull requests.  I should have seen these issues before the first pull request.  If scistarter actually gave me their docs instead of only answering questions via email this would not have been a problem.

faustbrian:
No problem. Have to deal with bad documented services every day so know the feeling.

PR #79, opened 2017-06-13
HubSpot provider

swilla:


faustbrian:
@swilla Also needs to be added to https://github.com/SocialiteProviders/Providers/blob/master/build/split.sh

PR #88, opened 2017-07-19
Always use Arr::get.

lucasmichot:
Always use `Arr::get`
As `array_get` is an application helper that does nothing more than calling `Arr::get`

### DIFF
--- a/src/Buffer/Provider.php
+++ b/src/Buffer/Provider.php
@@ -2,6 +2,7 @@
 
 namespace SocialiteProviders\Buffer;
 
+use Illuminate\Support\Arr;
 use SocialiteProviders\Manager\OAuth2\User;
 use Laravel\Socialite\Two\ProviderInterface;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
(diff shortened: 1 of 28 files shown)


faustbrian:
Is there any benefit to using `Arr::get` instead of `array_get` I am missing?

lucasmichot:
Yes. You call the function you need instead of a calling a helper that call the same function
Which is a best practice in term of performance.
For instance `array_get` is never used in `laravel/framework`

lucasmichot:
https://github.com/SocialiteProviders/Providers/pull/88/commits/6bc5262c58095b1a528942f584d505ff576a4e55 Cannot understand why this time imports needs to be sorted by length

lucasmichot:
Ah OK for this repo we use laravel syntax

faustbrian:
Merged it for now even though I don't see to much of a benefit of this as it makes sense for Laravels Core to not use those methods as you want to keep the Framework fast and decoupled from helpers. In real world applications using the helpers would have no impact and it keeps code clean.

PR #99, opened 2017-08-29
Add ArcGIS Provider

davidpiesse:
Inclusion of ArcGIS OAuth2 provider. Also allows for self hosted Protal for ArcGIS instances.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ composer.lock
 phpunit.xml
 vendor
 
+.idea/modules.xml
+.idea/providers.iml
+.idea/vcs.xml
+.idea/workspace.xml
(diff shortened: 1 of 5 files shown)


faustbrian:
Also has to be added to https://github.com/SocialiteProviders/Providers/blob/master/build/split.sh

.gitignore
```diff
@@ -2,3 +2,7 @@ composer.lock
 phpunit.xml
 vendor
```
faustbrian: The .gitignore shouldn't be modified in any way

davidpiesse: Argh! wasn't meant to be included in the commit! - Bloody PHPStorm!

PR #101, opened 2017-09-05
Add Withings Provider, using Paxx/Withings's code

Alex131089:
Add Withings Provider, using Paxx/Withings's (https://github.com/Zn4rK/php-withings) code

Dev'd in https://github.com/Alex131089/socialiteproviders-withing


faustbrian:
License has to be exactly the same as every other provider, MIT.

faustbrian:
Will add it to packagist in an hour or two.

PR #103, opened 2017-09-19
Just set all attributes provided by OAuth Server.

jorgenb:
@faustbrian Can you please merge this?


faustbrian:
This will be a breaking change as the returned data will be invalid for current users. Is there any reason for this change?

jorgenb:
The original design is inherently flawed. The attributes being parsed in `mapUserToObject()` might not be returned by the Oauth 2 server at all for this provider and in some cases there might be additional attributes returned based on the requested scopes.

faustbrian:
Release `v2.1.0` should be available on packagist.

PR #104, opened 2017-09-25
Add expires in to VK provider

spyric:
Now there is no way to get Expires in property for Vkontakte driver. This PR fix it


faustbrian:
Will tag a release later today.

PR #108, opened 2017-11-13
Error 48001 when using "snsapi_base" to configure the value of the "s…

nanyi:
Fix weixin provider. when using "snsapi_base" to configure the value of the "scopes" parameter, throw exception "error 48001".


faustbrian:
Some details about what this fixes would be helpful.

PR #114, opened 2017-11-18
Add Okta Provider

chaseconey:


chaseconey:
This one is a little bit different because it requires a URL custom to your company. Not sure if just creating another config/env setting for that is ok or not. Please advise.

src/Okta/Provider.php
```diff
@@ -0,0 +1,105 @@
...
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    protected function getOktaUrl()
+    {
+        return config('services.okta.url');
```
faustbrian: Use https://github.com/SocialiteProviders/Manager/blob/master/src/ConfigTrait.php#L46-L52 for additional configuration keys.

chaseconey:
@faustbrian I think that should do it. Let me know what you think.

faustbrian:
Send a PR to https://github.com/SocialiteProviders/socialiteproviders.github.io if you want to add it to the docs.

PR #116, opened 2017-11-28
Add smashcast and remove hitbox

valh1996:
Hello, the hitbox service has been renamed smashcast, here is a PR with the necessary modifications.

The old provider hitbox wasn't working anymore.


faustbrian:
Please send a PR for a new Provider as the old one should be kept for reference and will be marked as abandoned on packagist.

PR #119, opened 2017-11-30
Fixed NPE and JSON for Flickr provider

Axieum:
Where the original would use a 'json()' method on the request, this was undefined. Hence, it now takes advantage of the 'json_decode()' function. Furthermore, the response I received did not contain a 'realname' entry and thus was throwing an NPE, a check is now in place to check for this field.

### DIFF
--- a/src/Flickr/Server.php
+++ b/src/Flickr/Server.php
@@ -47,6 +47,7 @@ class Server extends BaseServer
     {
         $data = $this->getProfile($data['user']['id']);
         $data = $data['person'];
+        $data['realname']['_content'] = isset($data['realname']['_content']) ? $data['realname']['_content'] : '';
 
         $user = new User();
         $user->id = $data['id'];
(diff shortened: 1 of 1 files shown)


faustbrian:
Please use array_get instead of the isset if else.

Axieum:
Would this be acceptable?
```php
$data['realname']['_content'] = array_get($data['realname']['_content']) ?: '';
```

I can't find any documentation on this `array_get()` function. Sorry!

faustbrian:
https://laravel.com/docs/5.5/helpers#method-array-get

Axieum:
Ah, cheers! I was searching PHP's documentation for the function.

faustbrian:
You can just do `array_get($data, 'realname._content');`, the last parameter will default to null.

madbob:
This fix has never been "released", version 3.0.0 of Flickr's provider do not includes this.
Is it possible to publish a new release?

Issue #128, opened 2017-12-12
Merge of Microsoft-Graph and Microsoft-Live into new Microsoft provider


Jimbolino:
Both providers use the same urls for requesting tokens and api calls:

https://github.com/SocialiteProviders/Providers/blob/master/src/Microsoft-Graph/Provider.php#L33
https://github.com/SocialiteProviders/Providers/blob/master/src/Microsoft-Live/Provider.php#L32

The Microsoft-Azure provider uses the deprecated V1.0 api, but i do like the configurable "tenant"
https://github.com/SocialiteProviders/Providers/blob/master/src/Microsoft-Azure/Provider.php#L36

Would it be ok if I:
- Create a new provider "microsoft"
- Mark these 3 as deprecated

faustbrian:
If Microsoft now uses one API for all those services than you can make a new provider.

PR #130, opened 2017-12-18
added new Microsoft provider

Jimbolino:
as discussed in #128
im not sure how the git subsplit works, so i didnt update that script.


faustbrian:
Please add `split Microsoft                src/Microsoft:git@github.com:SocialiteProviders/Microsoft.git                         "master"` to the build script.

Paulsky:
@Jimbolino Could you please add documentation in this project: https://github.com/SocialiteProviders/Documentation ? Because I would like to fix: https://github.com/SocialiteProviders/Documentation/issues/68

PR #144, opened 2018-02-14
Steem OAuth2 Provider for Laravel Socialite

frnxstd:
Steem Connect API v2 provider implemented


faustbrian:
Needs to be added to https://github.com/SocialiteProviders/Providers/blob/master/build/split.sh.

PR #147, opened 2018-02-19
Add IFSP OAUTH

douglasanpa:
Add IFSP OAUTH


faustbrian:
Also needs to be added to https://github.com/SocialiteProviders/Providers/blob/master/build/split.sh

douglasanpa:
done

PR #158, opened 2018-03-21
fix Kakao user properties

kimbeejay:


faustbrian:
Why would you move the nickname to the name instead of the nickname field like it is currently?

kimbeejay:
Actually Kakao does not have real name field and uses nickname instead of it

faustbrian:
Then the nickname should be used for name and nickname because this will otherwise be a breaking change.

kimbeejay:
Hm.. indeed. I'll fix this

PR #160, opened 2018-03-29
Update to work with the new Twitch API

SnakeME:
Twitch has published a new api. The endpoints and the headers for authorization of the API have changed.

https://dev.twitch.tv/docs/authentication/#sending-user-access-and-app-access-tokens
https://dev.twitch.tv/docs/api/


faustbrian:
Is the old API gone and no longer works? Asking because I want to know if this is a breaking change.

PR #177, opened 2018-05-30
Update API URL and JWT

kailichiang:
1. Update LINE API URLs
2. Add "space" scopeSeparator
3. Add default scopes
4. Add JWT and get user from JWT

### DIFF
--- a/src/Line/Provider.php
+++ b/src/Line/Provider.php
@@ -14,6 +14,24 @@ class Provider extends AbstractProvider implements ProviderInterface
     const IDENTIFIER = 'LINE';
 
     /**
+     * The separating character for the requested scopes.
+     *
+     * @var string
+     */
+    protected $scopeSeparator = ' ';
+    
+    /**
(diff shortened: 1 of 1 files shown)


faustbrian:
Is this a completely new version of their API that is incompatible with the previous one?

kailichiang:
I'm not familiar with the previous one. I can use the same client_key and client_secret to get user details, from either the previous API or the new API. But the new version (API v2.1) provides JWT and user email which is not supported in the previous version.

However, names in user details of response are different from JWT and token.
The mapping is as below

| Field   | From access token | From JWT  |
| ------ |-------------| -----|
| id | userId | sub |
| name | displayName      |   name|
| email | N/A      |    email |
| avatar | pictureUrl      |    picture |

Hope the information helpful.

faustbrian:
Tagged version 2.0.0 available at https://packagist.org/packages/socialiteproviders/line because this looks like a breaking changed.

PR #179, opened 2018-06-05
Change url to v2 and check keys exist

scottxxx666:
1. Chang api url to v2
2. Check keys exist
3. Check email is valid or verified

reference:
- https://developers.kakao.com/docs/restapi/user-management
- https://devtalk.kakao.com/t/v2-api/51366/2


faustbrian:
Please use https://laravel.com/docs/5.6/helpers#method-array-get.

PR #185, opened 2018-06-13
Add Odnoklassniki provider

talgat065:


src/Odnoklassniki/composer.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "socialiteproviders/odnoklassniki",
+    "description": "Odnoklassniki OAuth2 Provider for Laravel Socialite",
+    "license": "MIT",
+    "authors": [{
+        "name": "Brian Faust",
```
faustbrian: This should be your name.

src/Odnoklassniki/composer.json
```diff
@@ -0,0 +1,18 @@
...
+        "email": "hello@brianfaust.de"
```
faustbrian: This should be your email.

.styleci.yml
```diff
@@ -0,0 +1 @@
+preset: symfony
```
faustbrian: This file is not needed.

PR #199, opened 2018-07-10
[Change] Introduces the Ability to Login to a Specific (Microsoft Graph) Tenant

Jaspaul:
This change was made to accommodate the ability to configure the tenant that you are authenticating against in the graph api. It covers the case where you would prefer to use the tenant specific profile for a guest user. I.E. All users authenticating to my application must be in the same tenant on Microsoft.

I discovered this issue when I attempted to add a guest user to my (multi-tenant) application. Since the application I'm working on maps it's tenants to Microsoft graph tenants I wanted to ensure that the guest user that was logging into my application does in fact have an account on the corresponding Microsoft tenant. With the common login endpoint the user was being returned with an id that did not match the user listing specific to a tenant. To resolve this I had to update to leverage the oauth endpoint for that specific tenant vs the common authentication endpoint.

The goal of this change was to ensure we preserve the existing functionality as the default behaviour so existing deployments leverage the common authentication endpoint.

### DIFF
--- a/src/Microsoft-Graph/Provider.php
+++ b/src/Microsoft-Graph/Provider.php
@@ -26,11 +26,45 @@ class Provider extends AbstractProvider implements ProviderInterface
     protected $scopeSeparator = ' ';
 
     /**
+     * Allows you to override the tenant id that the provider is configured
+     * with.
+     *
+     * @param string $tenantId
+     *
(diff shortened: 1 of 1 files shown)


m1guelpf:
@Jaspaul Could you also PR this to the docs repo?

PR #201, opened 2018-07-20
Add Mattermost oAuth2 Provider

takuya:
Hi , i add Mattermot instance OAuth2.


m1guelpf:
@takuya Could you PR the files instead of the submodule (we'll take care of the split internally), add your provider to the split.sh script & PR docs to the docs repo?

PR #209, opened 2018-09-04
Updated UFS provider to allow manager 2.0 or 3.0

claudsonm:
The composer.json file was updated to allow both versions of socialite manager.


m1guelpf:
@claudsonm Could you resolve conflicts?

PR #212, opened 2018-09-08
Added Facebook provider

autaut03:


m1guelpf:
Can you add this to the split.sh file & also PR it to the docs repo?

autaut03:
Here you go ;)

Note: as from what I see, because I actually extended a default socialiate provider and not the one socialiteProviders provides, `stateless()` method does not accept an argument, same as GooglePlus provider.

https://github.com/SocialiteProviders/Documentation/pull/11

PR #221, opened 2018-10-01
Mattermost moved from root to src folder

claudsonm:
Hi there,

I noticed that the mentioned provider were in the wrong place. This PR places it in the right location.
Another PR to the Docs repo is on the way.

Also, a few formatting stuff was made.


m1guelpf:
@claudsonm Thanks and sorry for this

PR #231, opened 2018-10-30
fixed Naver mapUserToObject

apollopy:
fixed notice error. The email and name fields may not be approved by naver.


apollopy:
Thanks, when can post to packagist.org？

PR #251, opened 2019-01-10
Fix broken PSR-4 autoloader

antonkomarev:
These providers were broken:
- Microsoft-TeamService
- App.net
- Acclaim
- UCL
- Teamleader

PSR-4 autoloader tried to find sources in not exists `/src` directory.

Additionally I've found that these 2 Providers don't have split repositories (but it exists in split.sh):
- UCL
- Teamleader


antonkomarev:
I've checked all `split.sh` file it contains all the repositories. But I haven't checked if all of them exists as split repositories on GitHub. Only those which I've fixed recently.

PR #253, opened 2019-01-10
Remove nested src folder to have consistent directories structure

antonkomarev:
This PR is a part of the https://github.com/SocialiteProviders/Manager/issues/132

It wouldn't be a breaking change because namespace wouldn't change, it will be just modified in local `composer.json` file and on next composer update it will just change this value in autoloader. After this change we will have completely identical directory structure all over the repository.


dingdayu:
When I submitted the PR, I noticed the problem with the `src` directory, which was different from other adapters. Because it was generated directly by the [generator](https://github.com/SocialiteProviders/Generators), I also specifically asked the PR.

Should I update the  [generator](https://github.com/SocialiteProviders/Generators)?

m1guelpf:
@antonkomarev Yes please

PR #265, opened 2019-01-29
Fixes to Steam provider

autaut03:
Hey. I made a few changes to Steam provider for it to be easier to both test and use:
- changed private to protected, to allow to overwrite the method if needed
- changed direct config references (to 'redirect' and 'client_secret') to corresponding class properties, so they can be set using socialite's default methods. That's how it works in other providers
- overwritten additionalConfigKeys method to allow the use of 'proxy' config value

All of the changes should be safe. I would be thankful if you could merge and release the changes as soon as possible, because there's no way I know of to pull from forked splitted repo.

### DIFF
--- a/src/Steam/Provider.php
+++ b/src/Steam/Provider.php
@@ -81,10 +81,8 @@ class Provider extends AbstractProvider
      */
     protected function getUserByToken($token)
     {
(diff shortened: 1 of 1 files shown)


autaut03:
Thanks a lot!

PR #266, opened 2019-01-30
Added Google Sign-in

xstoop:


m1guelpf:
Please add to docs

xstoop:
Ok,I didn't found need add docs  in the contribute page https://socialiteproviders.netlify.com/contribute.html.

Otherwise i have a question,if I need to remove the Google+ providers? Google+ will shut down soon .In Google+  provider class use the `const IDENTIFIER='google'` and `namespace SocialiteProviders\Google`, but it using the Google-Plus dirname. In this PT,i just created a new file to add Google sign-in and using the `const IDENTIFIER='google'` and `namespace SocialiteProviders\Google` same of the google+ provider.

About the doc, the google+ provider's doc has used the `google.md` filename, but the  title in the doc is `google+`.So i need to edit this `google.md` content or i just change the `google.md` name to `google+.md` and create a new `google.md` file?

How can i do ?

src/Google/Provider.php
```diff
@@ -0,0 +1,86 @@
...
+        return 'https://oauth2.googleapis.com/token';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getUserByToken($token)
+    {
+        $response = $this->getHttpClient()->get(
+            'https://www.googleapis.com/plus/v1/people/me', [
```
CSNWEB: This seems to be the old google plus api that is going to be deprecated soon?

src/Google/Provider.php
```diff
@@ -0,0 +1,86 @@
...
+use Illuminate\Support\Arr;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    const IDENTIFIER = 'GOOGLE';
```
CSNWEB:
```suggestion
    const IDENTIFIER = 'GOOGLEOAUTH';
```

src/Google/Provider.php
CSNWEB:
```suggestion
            'https://www.googleapis.com/oauth2/v3/userinfo', [
```
Got it working with this and the mapping changes below

xstoop:
@m1guelpf https://github.com/SocialiteProviders/Documentation/pull/34

src/Google/Provider.php
```diff
@@ -0,0 +1,86 @@
...
+use Illuminate\Support\Arr;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider implements ProviderInterface
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    const IDENTIFIER = 'GOOGLEOAUTH';
```
antonkomarev: Why it doesn't follow same convention with another providers? It should be:
```
const IDENTIFIER = 'GOOGLE';
```

antonkomarev: Ah... I see.. that's because GooglePlus has wrong identifier :\

Issue #267, opened 2019-02-01
Multiple scopes aren't working on makerlog


spekulatius:
When defining more than one scope I get a 405:

![image](https://user-images.githubusercontent.com/8433587/52145596-fe4cba00-2693-11e9-9ba6-a20b8733a061.png)

in the address bar it says:

    http://127.0.0.1:8000/makerlog/callback/?error=invalid_scope&state=....

Code:

`
        /**
         * @param string $driver
         * @return redirect
         */
        public function redirectToSocial($driver)
        {
            return Socialite::driver($driver)->scopes([
                'tasks:read',
                'tasks:write',
            ])->redirect();
        }

        /**
         * @param string $driver
         * @return redirect
         */
        public function handleSocialCallback($driver)
        {
                // the social media information
                $social_user = Socialite::driver($driver)->user();
        }
`

spekulatius:
Resolved and merged

PR #268, opened 2019-02-01
MINOR: fixes multiple scopes on makerlog #276

spekulatius:


spekulatius:
This should fix #267. Maybe @m1guelpf could verify this?

PR #271, opened 2019-02-07
Replace Plus API with People API due Plus shutdown

drfenixion:
Needs to activate Google People API (in Google console)

### DIFF
--- a/src/Google-Plus/Provider.php
+++ b/src/Google-Plus/Provider.php
@@ -2,11 +2,11 @@
 
 namespace SocialiteProviders\Google;
 
-use Illuminate\Support\Arr;
-use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
+use Laravel\Socialite\Two\ProviderInterface;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 
-class Provider extends AbstractProvider
+class Provider extends AbstractProvider implements ProviderInterface
 {
     /**
      * Unique Provider Identifier.
@@ -17,9 +17,8 @@ class Provider extends AbstractProvider
      * {@inheritdoc}
      */
     protected $scopes = [
-        'https://www.googleapis.com/auth/plus.me',
-        'https://www.googleapis.com/auth/plus.login',
-        'https://www.googleapis.com/auth/plus.profile.emails.read',
+        'profile',
+        'email',
     ];
 
     /**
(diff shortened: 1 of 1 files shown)


src/Google-Plus/Provider.php
```diff
@@ -51,10 +50,11 @@ protected function getTokenUrl()
     protected function getUserByToken($token)
     {
         $response = $this->getHttpClient()->get(
-            'https://www.googleapis.com/plus/v1/people/me', [
+            'https://people.googleapis.com/v1/people/me', [
             'headers' => [
                 'Authorization' => 'Bearer '.$token,
             ],
+            'query'   => array('personFields' => 'emailAddresses,names,photos'),
```
m1guelpf: Please use the new array style

src/Google-Plus/Provider.php
```diff
@@ -66,10 +66,11 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'     => $user['id'], 'nickname' => Arr::get($user, 'nickname'),
-            'name'   => $user['displayName'],
-            'email'  => $user['emails'][0]['value'],
-            'avatar' => Arr::get($user, 'image.url'),
+            'id' => $user['names'][0]['metadata']['source']['id'],
+            'nickname' => $user['names'][0]['displayName'],
```
m1guelpf: Are all of this values always set?

src/Google-Plus/Provider.php
```diff
@@ -66,10 +66,11 @@ protected function getUserByToken($token)
...
+            'name' => $user['names'][0]['displayName'],
+            'email' => $user['emailAddresses'][0]['value'],
+            'avatar' => $user['photos'][0]['url'],
```
m1guelpf: Is the first photo always the avatar? What happens if the user has no avatar?

drfenixion: If not set any ava Google uses an image with your name first char as the avatar. I removed my ava and  Google automatically set this ava for me https://lh6.googleusercontent.com/-IPi1e97NLh4/AAAAAAAAAAI/AAAAAAAAAAA/ACevoQNXgMp-CZ4U_4r_aIDPCcDJil4N_g/s100/photo.jpg

src/Google-Plus/Provider.php
drfenixion: Not found info about it. Due to this add some checks via Arr:get

Docs and test tool
https://developers.google.com/people/api/rest/v1/people/get?apix_params=%7B%22resourceName%22%3A%22people%2Fme%22%2C%22personFields%22%3A%22photos%22%7D

src/Google-Plus/Provider.php
drfenixion: Done

PR #283, opened 2019-02-18
Switched to api v2

throttlehead:
So v2 is LinkedIn's API is far more convoluted.

I ended up having to do two requests. A separate request needed to be done for the primary email address. Here are some pertinent docs

https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/profile-api?context=linkedin/compliance/context

https://docs.microsoft.com/en-us/linkedin/shared/integrations/people/primary-contact-api?context=linkedin/compliance/context

Getting a usable avatar URL was also much more tricky. I had to add a URN id which returns a collection of usable avatars of varying sizes. I ended up popping off the last one, which is the largest size available.

This could probably use some array key checking... this API is heavily nested.

### DIFF
--- a/src/LinkedIn/Provider.php
+++ b/src/LinkedIn/Provider.php
@@ -4,6 +4,7 @@ namespace SocialiteProviders\LinkedIn;
 
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
+use Illuminate\Support\Collection;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
 use Symfony\Component\HttpFoundation\RedirectResponse;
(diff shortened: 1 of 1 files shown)


m1guelpf:
Have you tested this?

throttlehead:
We are currently testing our LinkedIn signup/login with this update now. My own tests were successful.

I'll let you know how it goes.

throttlehead:
So far I've seen an error where profilePicture was not set on the response array. I'm assuming this was on a new account they set up for testing and the api might not return the key if a user does not have an avatar. I've added an array_key_exists check to prevent the error.

m1guelpf:
Then is this ready for merge?

throttlehead:
I'd say it should be safe to merge, but has this project been superseded by this provider in the socialite project itself? It looks like they ended up doing a similar pattern for this:

https://github.com/laravel/socialite/blob/4.0/src/Two/LinkedInProvider.php

m1guelpf:
Will tag a new version later today

PR #303, opened 2019-04-27
Add ClaveUnica driver

rodrigore:


rodrigore:
Hi, I already added to the documentation repo https://github.com/SocialiteProviders/Documentation/pull/46

rodrigore:
Hi Miguel, how can I test the package in my project? I mean, when I try to install the package:
`composer require socialiteproviders/claveunica`

It seems that it cannot find on packagist. Is that the problem?

PR #333, opened 2019-08-19
Some array keys may be undefined for Microsoft Graph

harrygulliford:
This was originally proposed in the read-only split of the Microsoft Graph repo [here](https://github.com/SocialiteProviders/Microsoft-Graph/pull/3), but the user who created the PR no longer exists.

From that PR...

> For example, Outlook user is not obligated to fill out the given name field, in that case we don't get response with givenName value therefore givenName index is undefined.
> 
> Similar situation was with Github provider, following commit [laravel/socialite@38c15dc](https://github.com/laravel/socialite/commit/38c15dc9ec9cd0132e45b3228e20eed8832bf9f3) set null as email index value if email is undefined.

In addition to that PR, I have changed from `array_get` to `Arr::get()` because the global array helpers are deprecated in 5.8 and will be removed in 6.0

### DIFF
--- a/src/Microsoft-Graph/Provider.php
+++ b/src/Microsoft-Graph/Provider.php
@@ -2,6 +2,7 @@
 
 namespace SocialiteProviders\Graph;
 
+use Illuminate\Support\Arr;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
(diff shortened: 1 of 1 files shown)


joearcher:
Please can we get this merged? This package is currently useless to me without these changes.

PR #336, opened 2019-08-29
Ensure that ext-json is also loaded.

lucasmichot:
Ensure that the JSON extension is also loaded.


m1guelpf:
@lucasmichot Could you also PR this to the generators repo?

lucasmichot:
Sure @m1guelpf , see https://github.com/SocialiteProviders/Generators/pull/24

PR #343, opened 2019-09-05
Fix properties field may be empty of Kakao user

### LINKED ISSUE #342: Kakao user, may be empty properties field.
I found that the problem occurred in certain Kakao users.   

`Undefined index: properties`

[see](https://devtalk.kakao.com/t/properties/53284/2)(This is written in Korean.)  

The article says that accounts that are not associated with any Kakao service may not have a 'properties' field.  

In fact I was able to find the response in my log in the following format:
```json
{"id":1,"kakao_account":{"has_email":true,"email_needs_agreement":false,"is_email_valid":true,"is_email_verified":true,"email":"example@example.com"}}
```

Even when the 'properties' field does not exist, it needs to be changed to operate normally.

cherryred5959:
closed #342  

This PR has been changed to work properly even if the 'properties' field of Kakao user is empty.

### DIFF
--- a/src/Kakao/KakaoProvider.php
+++ b/src/Kakao/KakaoProvider.php
@@ -98,8 +98,8 @@ class KakaoProvider extends AbstractProvider
 
         return (new User())->setRaw($user)->map([
             'id'        => $user['id'],
-            'nickname'  => $user['properties']['nickname'],
-            'name'      => $user['properties']['nickname'],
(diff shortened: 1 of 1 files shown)


cherryred5959:
@m1guelpf Thank you for merging, I hope to have new tags released to your library soon!

cherryred5959:
@m1guelpf plz, apply the merge to the https://github.com/SocialiteProviders/Kakao package.

Issue #344, opened 2019-09-07
Discord scopes are explicitly defined in the provider


jimsug:
The Discord Provider overrides the default (empty) array of scopes. Because of the way that `scopes()` is implemented on `BaseProvider`, this means that if developers don't want to request email information, passing `->scopes(['identify'])` (i.e. explicitly _not_ requesting email) doesn't function as expected.

https://github.com/SocialiteProviders/Providers/blob/master/src/Discord/Provider.php#L18

It is unclear from the contribution guidelines whether you're accepting such changes, happy to fix it if you are.

m1guelpf:
Feel free to PR this :D

SeoFood:
Workaround:

```php
Socialite::with('discord')
  ->setScopes([])
  ->redirect();
```

jimsug:
> 
> 
> Workaround:
> 
> ```
> Socialite::with('discord')
>   ->setScopes([])
>   ->redirect();
> ```

Is this tested and works, though? I would expect that you must pass at least one valid scope to a provider.

The `setScopes()` methods does an `array_merge()`, and this means that that hardcoded and explicitly defined scopes will be merged with the empty array: https://github.com/laravel/socialite/blob/4.0/src/Two/AbstractProvider.php#L305

Even if it works, though it's probably the wrong behaviour to require.

jimsug:
Hrm. Having looked at the other providers, there seem to be default scopes set, so changing this would break the pattern. I don't think the PR is required.

mackhankins:
I don't want email forced.  Ugh.

PR #345, opened 2019-09-09
fix scopes; closes SocialiteProviders/Providers#344

jimsug:


jimsug:
As per comment on #344 - this would break a pattern of default scopes, so I'm closing this.

PR #360, opened 2019-11-05
Added Instagram Basic API

pmochine:
Since Instagram is changing the API, I needed to change some things. The tutorial is found here: https://developers.facebook.com/docs/instagram-basic-display-api

I tested it on my server and it worked:

1. Installation: 

```
    // it's not added to packagist, someone needs to do it I guess? 
    //However, in future it's going to be:
    composer require socialiteproviders/instagram-basic
```

3. For the listener:

        'SocialiteProviders\Manager\SocialiteWasCalled' => [
            'SocialiteProviders\InstagramBasic\InstagramBasicExtendSocialite@handle',
        ],

4. Configuration setup

```
    'instagrambasic' => [
        'client_id' => env('INSTAGRAM_BASIC_KEY'),
        'client_secret' => env('INSTAGRAM_BASIC_SECRET'),
        'redirect' => env('INSTAGRAM_BASIC_REDIRECT'),
    ],
```

5. Usage

`     return Socialite::with('Instagrambasic')->redirect();`


atymic:
From the docs:

> Data returned by the API cannot be used to authenticate your app users or log them into your app. If you need an authentication solution we recommend using Facebook Login instead.

I was looking into this and it seems sign in with instagram is being deprecated and this isn't allowed to be used for that.

pmochine:
@atymic 😱I didn't see that! OH NO. But what now? I still need the option to connect to Instagram to fetch the user's feed. 

And I would still use Socialite. However, I understand when the core value of this plugin is to log in a user.

mpavao:
your addition still make sense (for validating and pulling timeline data), but user the experience needs to be more "on demand" than passive.

atymic:
I think it's ok to add :)

Just worth making it clear in the docs that it's not meant to be used for standard sign ins :)

pmochine:
Big news everyone. Instagram added long-lived access tokens. However, only for 30 days.

https://developers.facebook.com/docs/instagram-basic-display-api/guides/long-lived-access-tokens

pmochine:
@m1guelpf 
> Please open and link to a PR on the Documentation repository.

https://github.com/SocialiteProviders/Documentation/pull/79

pmochine:
@m1guelpf any news?

Te7a-Houdini:
any updates on this PR ? @m1guelpf

pmayet:
@m1guelpf It's ok to merge ?

src/Instagram-Basic/Provider.php
```diff
@@ -0,0 +1,124 @@
...
+
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    const IDENTIFIER = 'INSTAGRAMBASIC';
```
atymic, line 13:
```suggestion
    public const IDENTIFIER = 'INSTAGRAMBASIC';
```

src/Instagram-Basic/Provider.php
```diff
@@ -0,0 +1,124 @@
...
+     */
+    protected function mapUserToObject(array $user)
+    {
+        return (new User())->setRaw($user)->map([
+            'id' => $user['id'],
+            'nickname' => $user['username'],
+            'name' => null,
+            'email' => null,
+            'avatar' => null,
+            'account_type' => $user['account_type'],
```
atymic, line 78: Maybe worth using `Arr::get()` in case of upstream changes

radicalloop:
@atymic Waiting for to merge this PR.

src/Instagram-Basic/Provider.php
pmochine: Not really sure why `Arr::get()`. Do you mean if the key `account_type` has been changed that we get null instead of an error?

pmochine:
@atymic Rendamed InstagramExtendSocialite.php to InstagramBasicExtendSocialite.php ( I think that must be ok)

atymic:
Looks good. Please fix the style issues, and add it to the monorepo split file here:
https://github.com/SocialiteProviders/Providers/blob/master/monorepo-builder.yml
Then I'll merge/release.

Repo is here:
https://github.com/SocialiteProviders/InstagramBasic

pmochine:
@atymic have another look. What should I do now?

src/Instagram-Basic/Provider.php
```diff
@@ -0,0 +1,123 @@
+<?php
```
atymic, line 1: Sorry, just noticed the folder name does not comply with PSR-4. Could you rename it to `InstagramBasic`? (will add a CI step for this as well)

pmochine: Done

pmochine:
@atymic done

atymic:
https://packagist.org/packages/socialiteproviders/instagram-basic
Released and tagged :)

Minor issue on the docs, once you fix that I will merge and add to the site!

PR #363, opened 2019-11-08
fix: fix automatic splitting failure

atymic:
After a ton of debugging, I finally figured out why the splitting doesn't work in actions. When actions checks out a branch, it ends up on a detached head version of that branch. Plenty of others have issues with this as well: https://github.com/actions/checkout/issues/6

I'm leaving this as a draft for the moment as currently you need to supply an access token that has write permissions to all of the target repos. If the deploy fails, it ends up in the logs (not good!).

I'll have a go at switching it to use an SSH key (but then the CI will have access to any private repos of the person who creates it) but ideally we'd have a better way to ensure it remains secret, so will look into that.

cc @m1guelpf

### DIFF
--- a/.github/workflows/monorepo-split.yml
+++ b/.github/workflows/monorepo-split.yml
@@ -12,13 +12,21 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v1
+
+      # Required due to GH detached head bug https://github.com/actions/checkout/issues/6
+      - name: Prepare repository
+        run: git checkout "${GITHUB_REF:11}"
+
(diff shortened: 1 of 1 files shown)


atymic:
Upstream PR to the monorepo project to hide secrets from logs:
https://github.com/Symplify/MonorepoBuilder/pull/6

atymic:
Idea on permissions: What if we create a github user and add them to the providers ORG, giving them access to all the repos. We can then add an SSH key to that user,  and use that in the GH action to push to the other repos.

atymic:
ping @m1guelpf

PR #373, opened 2019-12-03
Add Auth0 socialite provider

chaseconey:


chaseconey:
@m1guelpf Should be good to go!

chaseconey:
@m1guelpf Let me know if you see any changes that are needed. Thanks! :)

AndyWendt:
I don't have access to merge etc anymore so I'm afraid I won't be of much help

src/Auth0/Provider.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function mapUserToObject(array $user)
+    {
+        return (new User())->setRaw($user)->map([
+            'id' => $user['sub'],
+            'nickname' => $user['nickname'],
+            'name' => $user['given_name'] . ' ' . $user['family_name'],
```
atymic, line 81: Will these always be set? Not clear from the docs.

src/Auth0/Provider.php
```diff
@@ -0,0 +1,96 @@
...
+            'email' => $user['email'],
+           'avatar' => null,
```
atymic, line 83: indenting wonky here

src/Auth0/Provider.php
chaseconey: Everything I have seen says yes.

chaseconey:
Should be good to go.

src/Auth0/Provider.php
AndyWendt: You could use the null coalescing operator so that it doesn't error on accident: https://stackoverflow.com/questions/34571330/php-ternary-operator-vs-null-coalescing-operator

'name' => $user['given_name'] ?? '' . ' ' . $user['family_name'] ?? '',

I expect that it would be good to do for the other values here.

AndyWendt:
@chaseconey would you mind updating the PR with style ci changes?

src/Auth0/Provider.php
atymic: Alternatively you can use array get `Arr::get($user, 'given_name', '')`

chaseconey:
@AndyWendt sorry for the delay, too many notifications. Style fixes applied, fallbacks applied for potential api null values.

Anything else you need from me?

src/Auth0/composer.json
```diff
@@ -0,0 +1,19 @@
...
+    "description": "Auth0 OAuth2 Provider for Laravel Socialite",
+    "license": "MIT",
+    "authors": [{
+        "name": "Chase Coney",
+        "email": "chase.coney@gmail.com"
+    }],
+    "require": {
+        "php": ">=5.5.9",
+        "ext-json": "*",
+        "socialiteproviders/manager": "^2.0"
```
atymic, line 12: Can we also allow v3 here please.

chaseconey: done.

atymic:
Thanks!

PR #376, opened 2019-12-08
fix: fix google repo URL

atymic:
This fixes a failure in the splitting script because the repo isn't actually called `Google`, it is `Google-Plus`

```
 remote: Repository not found.                                                
  fatal: repository 'https://github.com/SocialiteProviders/Google.git/' not f  
  ound                                      
```


atymic:
@m1guelpf

PR #386, opened 2020-02-08
Getty Images in providers

bbashel1:
add Getty Images in provider


src/GettyImages/Provider.php
```diff
@@ -0,0 +1,92 @@
...
+   }
+
+   /**
+    * {@inheritdoc}
+    */
+   protected function mapUserToObject(array $user)
+   {
+       return (new User)->setRaw($user)->map([
+           'first_name' => $user['first_name'],
+           'last_name'     => $user['last_name'],
```
atymic, line 89: alignment here is a bit wonky.

bbashel1:
added docs
https://github.com/SocialiteProviders/Documentation/pull/86

bbashel1:
added to the split config

atymic:
Can you fix the style issues?

PR #392, opened 2020-03-24
[Asana ] Fixed error in the mapUserToObject function

### LINKED ISSUE #391: Issue with Asana User object
Asana has changed the User object structure from using *id* to *gid*, see more [here](https://forum.asana.com/t/asana-is-moving-to-string-ids-updated-with-revised-timeline/29340).

Therefor, this line needs to be changed to support *gid*
https://github.com/SocialiteProviders/Providers/blob/8b866102fc34d2c1450318025e268f858ec8a5db/src/Asana/Provider.php#L54

johandavidsen:
Asana has changed the property name from id to gid. This commit is a fixed for issue #391.
Fixes #391


atymic:
Thanks!

atymic:
Tagged
https://github.com/SocialiteProviders/Asana/tree/v3.1.0

Issue #402, opened 2020-04-26
Deprecation Notice: Class SocialiteProviders does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0.


Jimbolino:
There are quite a few classes that are in a wrong file / folder, and are not compatible with the PSR-4 standard. 

Composer 2.0 will prevent these classes from loading, so a Deprecation Notice is thrown right now when you create an authoritative autoloader with: `composer dump -a`

Most are easy fixes, for example renaming the folder or using proper capitalization on the FileName.

```
Deprecation Notice: Class SocialiteProviders\Live\LiveExtendSocialite located in ./src/Microsoft-Live/LiveExtendSocialite.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class SocialiteProviders\Live\Provider located in ./src/Microsoft-Live/Provider.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class SocialiteProviders\AppNet\AppNetExtendSocialite located in ./src/App.net/AppNetExtendSocialite.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class SocialiteProviders\AppNet\Provider located in ./src/App.net/Provider.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class SocialiteProviders\Bitly\BitlyExtendSocialite located in ./src/Bit.ly/BitlyExtendSocialite.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
[...]

PR #403, opened 2020-04-27
Fix Class does not comply with psr-4 autoloading standard

Jimbolino:
fixes: https://github.com/SocialiteProviders/Providers/issues/402
I've only renamed Folders, the split repositories should be the same.
So people relying on those should be safe.

- [ ] Release major deviantart version
- [ ] Release major mattermost version
- [ ] Release 23andme
- [ ] Release 37Signals
- [ ] Release 500px
- [ ] Release App.net
- [ ] Release Battle.net
- [ ] Release Bit.ly
- [ ] Release Human-API
- [ ] Release Mediacube
- [ ] Release Microsoft-Azure
- [ ] Release Microsoft-Graph
- [ ] Release Microsoft-Live
- [ ] Release Microsoft-TeamService
- [ ] Release PayPal-Sandbox
- [ ] Release Weixin-Web


src/Deviantart/Provider.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace SocialiteProviders\deviantART;
+namespace SocialiteProviders\Deviantart;
```
owenvoke, line 3: Should the DeviantArt classes be under a `DeviantArt` directory and namespace to match the site naming?
```suggestion
namespace SocialiteProviders\DeviantArt;
```

Jimbolino: I'm matching the namespace as defined in DeviantartExtendSocialite.php
Changing the namespace of that file, would probably break a couple of people who already use it: https://packagist.org/packages/socialiteproviders/deviantart

owenvoke: Ah, ok. 👍

Jimbolino:
all better now :)

src/Deviantart/Provider.php
atymic: Thanks. Let me check if there's any other major version bumps required.

src/Mattermost/MattermostExtendSocialite.php
```diff
@@ -4,7 +4,7 @@
 
 use SocialiteProviders\Manager\SocialiteWasCalled;
 
-class MatterMostExtendSocialite
+class MattermostExtendSocialite
```
atymic, line 7: Guess this will need a major version bump as well?

Jimbolino: nope, it was actually the "style fixes" commit that changed it, i'm simply changing it back:
https://github.com/SocialiteProviders/Providers/commits/master/src/Mattermost/MatterMostExtendSocialite.php

atymic: Ahh. good catch, I will check if any other ones were changed.

src/Deviantart/Provider.php
owenvoke: If you're doing a major version for DeviantArt, is it worth making it consistent with the actual site naming then? 🤨

atymic: Yep, feel free :)

atymic:
@Jimbolino did you want to fix deviant art still, or should I tag the new major version? :)

owenvoke:
Oh, that was actually a request from me. I think it would make sense to fix it, however I can't at the moment.

Jimbolino:
My preference is towards backwards compatibility.
So only change the name if it's really needed.
For example when a company changes its name, and new users might not recognize the old one.

atymic:
Ok, no worries.
Will tag a major release now then :)

PR #405, opened 2020-04-29
Added Sign in with Apple Socialite Provider

Ahilmurugesan:
We added the Sign in with Apple provider for laravel socialite.  
This package includes JWT signature verification. 
We also added the suggested package for generating and managing the Apple Client Secret. 

https://github.com/Ahilmurugesan/socialiteproviders-apple
https://github.com/Ahilmurugesan/socialite-apple-helper
https://github.com/Ahilmurugesan/socialite-apple


src/Apple/Provider.php
```diff
@@ -0,0 +1,206 @@
...
+            'client_id'     => $this->clientId,
+            'redirect_uri'  => $this->redirectUrl,
+            'scope'         => $this->formatScopes($this->getScopes(), $this->scopeSeparator),
+            'response_type' => 'code',
+            'response_mode' => 'form_post',
+        ];
+
+        if ($this->usesState()) {
+            $fields['state'] = $state;
+            $fields['nonce'] = strtotime('12:00:00').'-'.$state;
```
jmarcher, line 70: Why is this time here hardcoded?

src/Apple/Provider.php
```diff
@@ -0,0 +1,206 @@
...
+    public static function verify($jwt)
+    {
+        $signer = new Sha256();
+
+        $token = (new Parser())->parse((string) $jwt);
+
+        if ($token->getClaim('iss') !== self::URL) {
+            throw new InvalidStateException('Invalid Issuer', Response::HTTP_UNAUTHORIZED);
+        }
+        /*if ($token->getClaim('aud') !== config('services.apple.client_id')) {
```
jmarcher, line 128: Delete commented code.

src/Apple/.php_cs.cache
```diff
@@ -0,0 +1 @@
+{"php":"7.3.17-1+ubuntu16.04.1+deb.sury.org+1","version":"2.16.3:v2.16.3#83baf823a33a1cbd5416c8626935cf3f843c10b0","indent":"    ","lineEnding":"\n","rules":{"blank_line_after_namespace":true,"braces":true,"class_definition":true,"constant_case":true,"elseif":true,"function_declaration":true,"indentation_type":true,"line_ending":true,"lowercase_keywords":true,"method_argument_space":{"on_multiline":"ensure_fully_multiline"},"no_break_comment":true,"no_closing_tag":true,"no_spaces_after_function_name":true,"no_spaces_inside_parenthesis":true,"no_trailing_whitespace":true,"no_trailing_whitespace_in_comment":true,"single_blank_line_at_eof":true,"single_class_element_per_statement":{"elements":["property"]},"single_import_per_statement":true,"single_line_after_imports":true,"switch_case_semicolon_to_colon":true,"switch_case_space":true,"visibility_required":true,"encoding":true,"full_opening_tag":true},"hashes":{"Provider.php":2176872719,"AppleExtendSocialite.php":1272416806}}
```
jmarcher, line 1: This file is not needed

src/Apple/.styleci.yml
```diff
@@ -0,0 +1 @@
+preset: symfony
```
jmarcher, line 1: This file is also not needed

src/Apple/Provider.php
Ahilmurugesan: It was supposed to be a unique string different from state. Now we have changed it to the current timestamp

src/Apple/Provider.php
Ahilmurugesan: Removed the commented code

src/Apple/.php_cs.cache
Ahilmurugesan: Removed the file

src/Apple/.styleci.yml
Ahilmurugesan: Removed the file

src/Apple/Provider.php
```diff
@@ -0,0 +1,203 @@
...
+            'client_id'     => $this->clientId,
+            'redirect_uri'  => $this->redirectUrl,
+            'scope'         => $this->formatScopes($this->getScopes(), $this->scopeSeparator),
+            'response_type' => 'code',
+            'response_mode' => 'form_post',
+        ];
+
+        if ($this->usesState()) {
+            $fields['state'] = $state;
+            $fields['nonce'] = time().'-'.$state;
```
jmarcher, line 70: If this needs to be a unique string you could rely on `Illuminate\Support\Str::uuid()` it is a dependency of Socialite anyway.

Ahilmurugesan: Thank You @jmarcher. We replaced time with in-built `Illuminate\Support\Str::uuid()`.

Ultima2876:
Hello @Ahilmurugesan - is it possible to use this in my project before it is officially merged into `SocialiteProviders::master`?

vonec:
https://github.com/Ahilmurugesan/socialiteproviders-apple im already using this one

src/Apple/Provider.php
```diff
@@ -0,0 +1,204 @@
...
+        $token = (new Parser())->parse((string) $jwt);
+
+        if ($token->getClaim('iss') !== self::URL) {
+            throw new InvalidStateException('Invalid Issuer', Response::HTTP_UNAUTHORIZED);
+        }
+        if ($token->isExpired()) {
+            throw new InvalidStateException('Token Expired', Response::HTTP_UNAUTHORIZED);
+        }
+
+        $data = json_decode(file_get_contents(self::URL.'/auth/keys'), true);
```
atymic, line 133: Do these keys change?
At minimum, we should cache this

src/Apple/Provider.php
```diff
@@ -0,0 +1,204 @@
...
+            throw new InvalidStateException('Invalid JWT Signature');
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function user()
+    {
+        //Temporary fix to enable stateless
```
atymic, line 155: Can you explain this?

src/Apple/Provider.php
```diff
@@ -0,0 +1,204 @@
...
+        static::verify($token);
+        $claims = explode('.', $token)[1];
+
+        return json_decode(base64_decode($claims), true);
+    }
+
+    /**
+     * Verify apple jwt.
+     *
+     * @param $jwt
```
atymic, line 114: is this a string?

src/Apple/Provider.php
```diff
@@ -0,0 +1,204 @@
...
+            ->setRefreshToken(Arr::get($response, 'refresh_token'))
+            ->setExpiresIn(Arr::get($response, 'expires_in'));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function mapUserToObject(array $user)
+    {
+        if (request()->filled('user')) {
```
atymic, line 183: Why are we reading from the request here? Shouldn't this come from the JWT?

src/Apple/Provider.php
Ahilanm: Apple signs the JWT using the public keys. We are not sure if the keys would remain constant, Apple started with single public key and then started using two keys now. We can cache this for an hour assuming Apple will not refresh or change the public keys. If Apple changes the public keys the token will fail.

So should we cache this?

src/Apple/Provider.php
Ahilanm: Submitted the PR with changes

src/Apple/Provider.php
Ahilanm: According to the [documentation](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms), the available valid values for scope(name and email) are sent as a user JSON object in form_post.

src/Apple/Provider.php
Ahilanm: According to Apple, the valid scopes(name and email) are only available under the response mode `form_post`. So the Socialite Apple callback should be always pointed at a post route while doing this Apple callback is initiating the new session from Apple which is different from the current user session. We tried if this is a cors issue and it is not. Currently, the CSRF is disabled for the callback route. In order to support the stateful applications which depend on sessions, we wrote a workaround so that the session state variable is passed to apple and then taken back. This is a temporary workaround. Users can still validate the state by comparing the session state variable from the Apple session with the existing session.

src/Apple/Provider.php
Ahilmurugesan: Apple signs the JWT using the public keys. We are not sure if the keys would remain constant, Apple started with single public key and then started using two keys now. We can cache this for an hour assuming Apple will not refresh or change the public keys. If Apple changes the public keys the token will fail.

So should we cache this?

src/Apple/Provider.php
Ahilmurugesan: According to Apple, the valid scopes(name and email) are only available under the response mode form_post. So the Socialite Apple callback should be always pointed at a post route while doing this Apple callback is initiating the new session from Apple which is different from the current user session. We tried if this is a cors issue and it is not. Currently, the CSRF is disabled for the callback route. In order to support the stateful applications which depend on sessions, we wrote a workaround so that the session state variable is passed to apple and then taken back. This is a temporary workaround. Users can still validate the state by comparing the session state variable from the Apple session with the existing session.

src/Apple/Provider.php
Ahilmurugesan: Submitted the PR with changes

src/Apple/Provider.php
Ahilmurugesan: According to the documentation, the available valid values for scope(name and email) are sent as a user JSON object in form_post.

src/Apple/Provider.php
Ahilmurugesan: Currently enabled cache for 5 minutes

src/Apple/Provider.php
Ahilmurugesan: Currently, we have implemented state verification using the JWT nonce. Users can still verify the state by passing the session `state_verify`  variable.

atymic:
Launched & tagged, thanks @Ahilmurugesan for the detailed explainations!
https://packagist.org/packages/socialiteproviders/apple

PR #413, opened 2020-05-06
Change twitch user request according to new policy

### LINKED ISSUE #411: Twitch provider change rules
Since 1 May 2020 Twitch.tv changed a rule to request to API endpoints `https://api.twitch.tv/helix/users`. 
It need to send both: **Client ID and OAuth token.**

So now, this provider _does not work_!
Fix it please.

Katzen48:
Closes #411 

See https://discuss.dev.twitch.tv/t/requiring-oauth-for-helix-twitch-api-endpoints/23916


MatthewSH:
I just check this locally. It fixes my issues as well. When can this be pushed? It's holding me back from deploying my project to the production servers because Twitch login just doesn't work.

PR #423, opened 2020-05-21
Update QuickBooks Socialite Driver

lukecurtis93:
As of Dec 17 2019 Quickbooks deprecated their oAuth1 implementation:
https://developer.intuit.com/app/developer/qbo/docs/develop/authentication-and-authorization/oauth-1.0a

This PR migrates that functionality to their oAuth2 flow.


src/QuickBooks/Provider.php
```diff
@@ -2,25 +2,87 @@
...
+
     /**
      * Unique Provider Identifier.
      */
     const IDENTIFIER = 'QUICKBOOKS';
 
+    /**
+     * Sandbox endpoint for retrieving user info.
+     */
+    const USERSANDBOXENDPOINT = 'https://sandbox-accounts.platform.intuit.com/v1/openid_connect/userinfo';
```
atymic, line 22: can we get some underscores in here to seperate words please :)

src/QuickBooks/Provider.php
```diff
@@ -2,25 +2,87 @@
...
+    {
+        return 'https://oauth.platform.intuit.com/oauth2/v1/tokens/bearer';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getUserByToken($token)
+    {
+        $endpoint = $this->getConfig('app.env', 'local') == 'production' ? self::USERPRODUCTIONENDPOINT : self::USERSANDBOXENDPOINT;
```
atymic, line 55: can you split this onto multiple lines please?

atymic: Also, strict comparison please `===`.
I would also prefer to make this configurable rather than using the current env?

src/QuickBooks/Provider.php
lukecurtis93: no worries :)

src/QuickBooks/Provider.php
lukecurtis93: good idea - added 👍

src/QuickBooks/Provider.php
```diff
@@ -2,25 +2,88 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getUserByToken($token)
+    {
+        $endpoint = self::USER_PRODUCTION_ENDPOINT;
+
+        if (env('QUICKBOOKS_ENV') === 'development') {
```
atymic, line 54: can we add a config in the service's specific config please?
This will break if you run `artisan config:cache`

lukecurtis93: for sure - do I need to make an additional PR for the README regarding this additional configuration too?

atymic:
@lukecurtis93 that would be great, can you PR the docs repo please?
https://github.com/SocialiteProviders/documentation
Thanks!

PR #428, opened 2020-05-25
Add Telegram provider

Max13:
The Telegram connect workflow is, different from OAuth:

- You include a `<script>` tag pointing to a hosted script at `Telegram's`.
- The script shows a button, opening (on click) telegram login form
- When logged in, the user is redirected to the `redirect_url` with all the user data and a `SHA256-HMAC` to verify, in the query.

So there is no `Socialite::driver('telegram')->redirect()` and calling that will call `abort(404)` with an error message.

My provider is based on `SocialiteProviders\Manager\OAuth2\AbstractProvider`, and the only differences for the user are:

- When auto-discovery is available, `composer.json` declares the `Socialite` facade, as usually, other providers require Socialite to be unloaded.
- When auto-discovery, the user can just prevent `Socialite` to be discovered by adding these lines to the project's `composer.json`:

```
"extra": {
    "laravel": {
        "dont-discover": [
            "laravel/socialite"
        ]
    }
}
```

`Socialite::driver('telegram')->user()` woks as intended.


Max13:
5145ad5 fixes the behavior of `redirect()` aborting with 404.

It now simply shows the login button allone on a page. It's not intended to be used (but rather showing the button directly without calling `redirect()`), but just in case... It's there.

src/Telegram/Provider.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        return null;
+    }
+
+    public function getButton()
+    {
+        $botname = $this->config['bot'];
+        $callbackUrl = $this->config['redirect'];
+
+        return '<script async src="https://telegram.org/js/telegram-widget.js" data-telegram-login="'.$botname.'" data-size="large" data-userpic="false" data-auth-url="'.$callbackUrl.'" data-request-access="write"></script>';
```
atymic, line 56: could we use `sprintf` instead of concat here please?

src/Telegram/Provider.php
```diff
@@ -0,0 +1,120 @@
...
+     */
+    protected function getUserByToken($token)
+    {
+        return null;
+    }
+
+    public function getButton()
+    {
+        $botname = $this->config['bot'];
+        $callbackUrl = $this->config['redirect'];
```
atymic, line 54:
```suggestion
        $callbackUrl = $this->getConfig('redirect');
```

atymic:
@Max13 Happy to add this provider, could you make the corrections as per the comments :)

src/Telegram/Provider.php
Max13: Ok, solved in 0735f8d

src/Telegram/Provider.php
Max13: Can you tell me why `getConfig()`, here, would be preferable? I chose `config[]` for consistency

atymic: Because it's using the abstracted version, in case how we load/store the config changes.

src/Telegram/composer.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "extra": {
+        "laravel": {
+            "aliases": {
+                "Socialite": "Laravel\\Socialite\\Facades\\Socialite"
+            }
+        }
+    },
+    "require": {
+        "php": "^5.6 || ^7.0",
```
atymic, line 22: drop < 7.2 here

src/Telegram/composer.json
```diff
@@ -0,0 +1,25 @@
...
+        "socialiteproviders/manager": "~2.0 || ~3.0"
```
atymic, line 23: Drop v2 here

src/Telegram/Provider.php
Max13: Resolved in 296cea3

Max13:
No problem. Done

atymic:
Dunno what happened there, was still showing a conflict but I fixed it. Thanks for the PR :)

atymic:
@Max13 I added some docs but they probably don't make complete sense due to how this provider works. Could your PR any needed changes when you have a chance?
https://github.com/SocialiteProviders/Providers/blob/e466ad53791eb8e2b07d44176052f797d91ec7cd/src/Telegram/README.md

PR #429, opened 2020-05-27
[Microsoft] Get avatar.

danswiser:
The Microsoft Graph API does not provide a URL to the users avatar from what I can tell. Microsoft actually returns the full binary avatar at the endpoint 'https://graph.microsoft.com/v1.0/me/photo/$value'.

Docs: https://docs.microsoft.com/en-us/graph/api/profilephoto-get?view=graph-rest-1.0#get-the-photo

This change makes an additional request to get the binary data and save it as the 'avatar_base64' property. Since it doesn't look like any other providers return the actual avatar and just link to the avatar URL, I'm not sure how maintainers would prefer to name the property. The 'avatar' property is left null since there's no public avatar URL.

### DIFF
--- a/src/Microsoft/MicrosoftAvatar.php
+++ b/src/Microsoft/MicrosoftAvatar.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace SocialiteProviders\Microsoft;
+
+use GuzzleHttp\Psr7\Response;
+
+class MicrosoftAvatar
+{
+    /**
+     * The Guzzle Response object.
+     *
+     * @var \GuzzleHttp\Psr7\Response
+     */
+    public $response;
+
+    /**
+     * Set the response of the avatar.
+     *
+     * @param string $response
+     *
+     * @return $this
+     */
+    public function setResponse(Response $response)
+    {
+        $this->response = $response;
+
+        return $this;
+    }
+
+    /**
+     * Return the content type of the avatar.
+     *
+     * @return string
+     */
+    public function getContentType(): string
+    {
+        return $this->response->getHeader('content-type')[0];
+    }
+
+    /**
+     * Return the avatar in binary.
+     *
+     * @return binary
+     */
+    public function getContents()
+    {
+        return $this->response->getBody()->getContents();
+    }
+
+    /**
+     * Return the data URI formatted image.
+     *
+     * @return string
+     */
+    public function __toString(): string
+    {
+        return 'data:'.$this->getContentType().';base64,'.base64_encode($this->getContents());
+    }
+}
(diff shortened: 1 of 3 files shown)


atymic:
Perhaps we could return a data URI?
Also, should we put this behind a flag? That way only those who need it can turn it on 🤔

danswiser:
For my usage I wanted to save the avatar locally and generate a URL to it but for general use, I think it does make more sense to return a data URI.

Do any other providers have additional flags? If this is the only provider that will have a flag, it makes me think it should be done differently.

atymic:
> For my usage I wanted to save the avatar locally and generate a URL to it but for general use, I think it does make more sense to return a data URI.

Actually, thinking about it a data URI prob won't fit into most DB columns.

What about this instead:

We extend the `AbstractUser` object, and have a method on it called `getAvatar` which returns a stream/file response. That way, the user CAN get the avatar and save it if needed?

danswiser:
> We extend the AbstractUser object, and have a method on it called getAvatar which returns a stream/file response. That way, the user CAN get the avatar and save it if needed?

I added the following methods to `SocialiteProviders\Microsoft\Provider`:

`public function avatar()`
`protected function getAvatarByToken($token)`
`protected function mapAvatarToObject($response)`

Just like there is for the user object. I do think it would make more sense to have `->user()->getAvatar()` but I didn't see any Http client in the AbstractUser class and don't want to to drastically change this provider compared to all the others. Thoughts?

atymic:
Looks good to me.
Could you make a PR to the docs repo as well please? :)

danswiser:
So I found some problems when the requests were stateful.

When a provider calls the `hasInvalidState()` method like in [SocialiteProviders\Manager\OAuth2\AbstractProvider.php#L36](https://github.com/SocialiteProviders/Manager/blob/master/src/OAuth2/AbstractProvider.php#L36) the state data is pulled out of the session in [Laravel\Socialite\Two\AbstractProvider.php#L247](https://github.com/laravel/socialite/blob/4.0/src/Two/AbstractProvider.php#L247) and causes any future requests to fail. 

I looked over the `Laravel\Socialite\Two\User` class and saw the public `$token` property that I didn't see before. Since the token was accessible, I was actually able to extend the class and build a second Guzzle request to get the avatar data. Now it's as simple as `Socialite::driver($provider)->user()->getAvatar()` like we wanted.

I'll make a PR to the docs shortly.

danswiser:
[Docs PR](https://github.com/SocialiteProviders/Documentation/pull/97)

atymic:
Awesome job. Thanks!

PR #432, opened 2020-06-02
Update SciStarter scope separator

biancarosenbaum:
This PR replaces the default comma scope separator with spaces, as required by the SciStarter API.

### DIFF
--- a/src/SciStarter/Provider.php
+++ b/src/SciStarter/Provider.php
@@ -15,6 +15,13 @@ class Provider extends AbstractProvider implements ProviderInterface
     const IDENTIFIER = 'SCISTARTER';
 
     /**
+     * The separating character for the requested scopes.
+     *
+     * @var string
+     */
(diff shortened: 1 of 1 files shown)


biancarosenbaum:
> Do you have a link to the docs where that's stated? I had a look and couldn't find it.

Really good question. It is not explicitly stated in the API docs, but the examples given show a space separating the scopes.

[SciStarter Example, line 70](https://scistarter.org/openid-connect-example-source-code)

`# Requesting the openid scope is a protocol requirement
# Requesting the profile scope gives us read access to
# basic user information, if the user grants it.
# Requesting the participation scope gives us the
# right to add participation events to the user's
# record, if the user grants it.
"scope": "openid profile participation",`

atymic:
> Really good question. It is not explicitly stated in the API docs, but the examples given show a space separating the scopes.

Perfect, thanks for the PR!

PR #433, opened 2020-06-08
[steam] Fix "realm and return_to do not match" when using HTTP

ilumos:
Currently when not using HTTPS (for example in dev environment) Steam provides the error "realm and return_to do not match" causing OpenIDValidationException to be thrown.

This PR stops the hardcoding of HTTPS for the 'realm' parameter, and instead uses HTTPS when the original request was made using HTTPS.

This PR also adds the error from Steam into the OpenIDValidationException message.

Thanks!

### DIFF
--- a/src/Steam/Provider.php
+++ b/src/Steam/Provider.php
@@ -140,11 +140,17 @@ class Provider extends AbstractProvider
     {
         $realm = $this->getConfig('realm', $this->request->server('HTTP_HOST'));
 
+        if ($this->request->secure()) {
+            $realm = sprintf('https://%s', $realm);
+        } else {
+            $realm = sprintf('http://%s', $realm);
+        }
+
         $params = [
             'openid.ns'         => self::OPENID_NS,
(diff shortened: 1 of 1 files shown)


src/Steam/Provider.php
```diff
@@ -140,11 +146,17 @@ private function buildUrl()
...
+        } else {
+            $realm = sprintf('http://%s', $realm);
+        }
+
         $params = [
             'openid.ns'         => self::OPENID_NS,
             'openid.mode'       => 'checkid_setup',
             'openid.return_to'  => $this->redirectUrl,
-            'openid.realm'      => sprintf('https://%s', $realm),
+            'openid.realm'      => $realm,
```
atymic, line 159:
```suggestion
            'openid.realm'      => sprintf('%s://%s', $this->request->getScheme(), $realm),
```

Save using the conditional :)

ilumos: thanks!

ilumos:
Let me know what you think :)

PR #438, opened 2020-06-18
Apple: Switch Query Encoding type to RFC3986

RomainLanz:
Hey there! 👋 

The current encoding used to generate URL by the Apple Provider make it fails on mobile.

The scopes should be separated with a space, and the default encoding will use a `+`.


atymic:
Thanks!

PR #439, opened 2020-06-23
Ability to use proxies when requesting

sofwar:
With a heavy load, steam bans IP for a while.


src/Steam/Provider.php
```diff
@@ -205,6 +205,7 @@ public function getDefaultRequestOptions()
     {
         return [
             RequestOptions::FORM_PARAMS => $this->getParams(),
+            RequestOptions::PROXY => $this->getConfig('proxy')
```
atymic, line 208: I assume `null` for this works fine?

sofwar: he just ignores him

PR #445, opened 2020-07-01
Fix quickbooks provider scopes

ragulka:
This PR fixes Quickbooks provider scopes by defining the scopes as an array of strings instead of one long string and by defining the `scopeSeparator`. 

The current implementation is broken because the default `scopeSeparator` is a comma, which triggers an `invalid_scope` error with QuickBooks when trying to customize scopes with `->scopes()` or `->setScopes()`.

In addition, it removes the extra scopes (`com.intuit.quickbooks.accounting`, `phone`, `address`) that are not relevant to getting basic user details. The idea is to keep the default scopes to a minimum while still being useful. This follows the example of the core Socialite providers as well as most of the other extra providers in this repo.

Finally, there are a few code style tweaks as well.

### DIFF
--- a/src/QuickBooks/Provider.php
+++ b/src/QuickBooks/Provider.php
@@ -26,7 +26,12 @@ class Provider extends AbstractProvider implements ProviderInterface
     /**
      * {@inheritdoc}
      */
-    protected $scopes = ['com.intuit.quickbooks.accounting openid profile email phone address'];
+    protected $scopes = ['openid', 'profile', 'email'];
+
+    /**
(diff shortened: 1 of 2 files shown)


ragulka:
@atymic thanks for the merge! Any indication when this may be tagged for release? It's blocking a downstream release on an app that I'm working on.

atymic:
Apologies, tagged 👍 
https://github.com/SocialiteProviders/QuickBooks/releases/tag/v3.0.2

PR #449, opened 2020-07-07
Fix issue #448

### LINKED ISSUE #448: Missing email and Incomplete name in zoho provider
Zoho provider has email field missing.

Also, name field uses only first name instead of full name.

adwiv:
Fix #448 

- Adds email field 
- Updates name field to use full name.


atymic:
Thanks!

PR #450, opened 2020-07-08
Migrate to Yahoo Oauth 2.0 API

varbark:
Since OpenID2 will be EOLed on 6/30/20,
We need to migrate from OpenID2 to OpenID Connect API
https://developer.yahoo.com/oauth2/guide/OpenID2/

SocDir (Social Directory Web Service) also EOL date is June 30, #2020
https://developer.yahoo.com/oauth/social-directory-eol/

### DIFF
--- a/src/Yahoo/Provider.php
+++ b/src/Yahoo/Provider.php
@@ -23,6 +23,9 @@ class Provider extends AbstractProvider
      */
     protected function getAuthUrl($state)
     {
+        $this->scopes(['openid2']);
+        $parse_url = parse_url($this->redirectUrl);
+        $this->with(['openid2_realm' => $parse_url['scheme'] . '://' . $parse_url['host']]);
         return $this->buildAuthUrlFromBase('https://api.login.yahoo.com/oauth2/request_auth', $state);
     }
 
@@ -39,13 +42,13 @@ class Provider extends AbstractProvider
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get('https://social.yahooapis.com/v1/user/'.$this->xoauth_yahoo_guid.'/profile?format=json', [
+        $response = $this->getHttpClient()->get('https://api.login.yahoo.com/openid/v1/userinfo', [
             'headers' => [
                 'Authorization' => 'Bearer '.$token,
             ],
         ]);
 
-        return json_decode($response->getBody(), true)['profile'];
+        return json_decode($response->getBody(), true);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Yahoo/Provider.php
```diff
@@ -39,13 +42,13 @@ protected function getTokenUrl()
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get('https://social.yahooapis.com/v1/user/'.$this->xoauth_yahoo_guid.'/profile?format=json', [
```
lucasmichot, line 42: Since you don't use `xoauth_yahoo_guid` class property anymore, can you remove it?

src/Yahoo/Provider.php
```diff
@@ -84,4 +87,4 @@ protected function parseAccessToken($body)
 
         return Arr::get($body, 'access_token');
     }
-}
+}
```
lucasmichot, line 90: Why removing the EOL here?

src/Yahoo/Provider.php
```diff
@@ -57,11 +60,11 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'       => $user['guid'],
+            'id'       => $user['sub'],
             'nickname' => $user['nickname'],
```
lucasmichot, line 64: What if the nickname is null?
I would fallback to sub

varbark: ok, i will add it

src/Yahoo/Provider.php
```diff
@@ -23,6 +23,9 @@ class Provider extends AbstractProvider
      */
     protected function getAuthUrl($state)
     {
+        $this->scopes(['openid2']);
```
lucasmichot, line 26: It is not needed. Just set in the scopes class porperty.
Please see \SocialiteProviders\Zendesk\Provider::$scopes

src/Yahoo/Provider.php
```diff
@@ -23,6 +23,9 @@ class Provider extends AbstractProvider
...
+        $parse_url = parse_url($this->redirectUrl);
```
lucasmichot, line 27: But you don't need all part here?

lucasmichot: Please limit the components you retrieve: ` parse_url($this->redirectUrl, PHP_URL_SCHEME | PHP_URL_HOST);`

lucasmichot: Also `$parse_url`:  variables are not noted in snake_case in this project, but in camelCase

lucasmichot, line 27: `$parse_url` should be in camelCase

src/Yahoo/Provider.php
varbark: done

varbark:
@lucasmichot   PR has been updated, plz check , Thanks

src/Yahoo/Provider.php
```diff
@@ -57,11 +60,11 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'       => $user['guid'],
-            'nickname' => $user['nickname'],
-            'name'     => trim(sprintf('%s %s', Arr::get($user, 'givenName'), Arr::get($user, 'familyName'))),
-            'email'    => Arr::get($user, 'emails.0.handle'),
-            'avatar'   => Arr::get($user, 'image.imageUrl'),
+            'id'       => $user['sub'],
+            'nickname' => $user['nickname'] || $user['sub'],
```
lucasmichot, line 64:
```suggestion
            'nickname' => Arr::get($user, 'nickname', Arr::get($user, 'sub')),
```

varbark: done

varbark: Thanks for this suggestion, safer way to get data from array

atymic:
Thanks @varbark @lucasmichot 

Are the docs up to date?

atymic:
https://github.com/SocialiteProviders/Yahoo/releases/tag/v3.0.2

PR #453, opened 2020-07-09
Ensure the syntax is compatible with PHP < 7.

lucasmichot:
Some part of the code introduce PHP 7 or PHP 7.1 new syntax, which breaks the compatibility.

This PR ensures that PHP versions < 7 are supported, as stated in the composer.json files.

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -21,7 +21,7 @@ class Provider extends AbstractProvider implements ProviderInterface
      */
     const IDENTIFIER = 'APPLE';
 
-    private const URL = 'https://appleid.apple.com';
+    const URL = 'https://appleid.apple.com';
 
     /**
      * {@inheritdoc}
@@ -196,15 +196,17 @@ class Provider extends AbstractProvider implements ProviderInterface
      */
     protected function mapUserToObject(array $user)
     {
+        $fullName = '';
+
         if (request()->filled('user')) {
             $userRequest = json_decode(request('user'), true);
 
             if (array_key_exists('name', $userRequest)) {
                 $user['name'] = $userRequest['name'];
                 $fullName = trim(
-                    ($user['name']['firstName'] ?? '')
+                    Arr::get($user, 'name.firstName', '')
                     .' '
-                    .($user['name']['lastName'] ?? '')
+                    .Arr::get($user, 'name.lastName', '')
                 );
             }
         }
@@ -213,8 +215,8 @@ class Provider extends AbstractProvider implements ProviderInterface
             ->setRaw($user)
             ->map([
                 'id'    => $user['sub'],
-                'name'  => $fullName ?? null,
-                'email' => $user['email'] ?? null,
+                'name'  => $fullName !== '' ? $fullName : null,
+                'email' => Arr::get($user, 'email'),
             ]);
     }
 }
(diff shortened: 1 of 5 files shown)


src/Apple/Provider.php
```diff
@@ -196,15 +196,17 @@ public function user()
...
     {
+        $fullName = '';
+
         if (request()->filled('user')) {
             $userRequest = json_decode(request('user'), true);
 
             if (array_key_exists('name', $userRequest)) {
                 $user['name'] = $userRequest['name'];
                 $fullName = trim(
-                    ($user['name']['firstName'] ?? '')
```
atymic, line 205: null coalesce is in 7.0

src/Line/Provider.php
```diff
@@ -87,11 +88,11 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'       => $user['userId'] ?? $user['sub'] ?? null,
+            'id'       => Arr::get($user, 'userId', Arr::get($user, 'sub')),
```
atymic, line 91: as above

src/Microsoft/MicrosoftAvatar.php
```diff
@@ -32,7 +32,7 @@ public function setResponse(Response $response)
      *
      * @return string
      */
-    public function getContentType(): string
+    public function getContentType()
```
atymic, line 35: i'm pretty sure 7.0 supports this?

lucasmichot: It does not match composer dependencies constraints: https://github.com/SocialiteProviders/Providers/blob/master/src/Microsoft/composer.json#L10

This package is supposed to work with `PHP 5.6`

And it is not the case now

src/Line/Provider.php
lucasmichot: Check https://github.com/SocialiteProviders/Providers/blob/master/src/Line/composer.json#L10

src/Apple/Provider.php
lucasmichot: Same as other comments @atymic, all those providers are supposed to work with `PHP 5.6`...

Introducing PHP 7 or 7.1 specific syntax breaks this compatibility.

atymic: Ok, feel free to drop < 7 then. We really need to drop < 7.2 some time soon as well.

lucasmichot:
Hey @atymic, thanks for your feedback.
I am happy to provide a PR that drops support for PHP < 7

> We really need to drop < 7.2 some time soon as well.

Why not doing it now directly then?

Is this something part of your backlog (if so, I could not see it)
cc / @m1guelpf

atymic:
> I am happy to provide a PR that drops support for PHP < 7

I'd just drop it for the ones using syntax that doesn't work on lower versions, then when we drop < 7.2 we can do all of them. 

>Why not doing it now directly then?
> Is this something part of your backlog (if so, I could not see it)

Feel free to create a PR.  7.0 hasn't been supported for almost a two and a half years, and 7.1 for one and a half.
https://www.php.net/supported-versions.php

atymic:
@lucasmichot 

I've merged your other PRs, thanks for all the awesome work 😀 
How about you PR dropping support for < `7.2` and then I'll make a new major version release of all providers?

Feel free to suggest any other breaking changes while we're at it :)

PR #456, opened 2020-07-09
Fix user email in Microsoft Graph Provider

xvladxtremal:


atymic:
cc @xvladxtremal

stidges:
@atymic I've just checked this in my application (which uses the Microsoft provider, but it uses the same API endpoints underneath) that in a lot of cases the `mail` property is empty, and the `userPrincipalName` is the field that contains the user's email address. I'm not sure where this is documented though. The Microsoft provider already uses the `userPrincipalName` for its email address field too. I think we may even be able to merge most of the Microsoft providers into one, since they generally use the same endpoints

src/Graph/Provider.php
```diff
@@ -102,7 +102,7 @@ protected function mapUserToObject(array $user)
         return (new User())->setRaw($user)->map([
             'id'    => $user['id'],
             'name'  => $user['displayName'],
-            'email' => $user['mail'],
+            'email' => $user['userPrincipalName'],
```
atymic, line 105:
```suggestion
            'email' => $user['mail'] ?? $user['userPrincipalName'],
```

How about we check both? That way no behavior should change and it also fixes your case.

PR #464, opened 2020-07-15
Microsoft: use configured tenant in getTokenUrl method

stidges:
Currently the `getTokenUrl` method is hardcoded to use the `common` tenant instead of the one configured by the user. This caused the token retrieval to fail for me, as I restricted my app to only personal accounts which requires you to use the `consumer` 'tenant'. Respecting the configured tenant fixes this.

### DIFF
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -41,7 +41,7 @@ class Provider extends AbstractProvider
      */
     protected function getTokenUrl()
     {
-        return 'https://login.microsoftonline.com/common/oauth2/v2.0/token';
+        return 'https://login.microsoftonline.com/'.($this->config['tenant'] ?: 'common').'/oauth2/v2.0/token';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Microsoft/Provider.php
```diff
@@ -41,7 +41,7 @@ protected function getAuthUrl($state)
      */
     protected function getTokenUrl()
     {
-        return 'https://login.microsoftonline.com/common/oauth2/v2.0/token';
+        return 'https://login.microsoftonline.com/'.($this->config['tenant'] ?: 'common').'/oauth2/v2.0/token';
```
atymic, line 44: Could we use a `sprintf` here please

stidges: @atymic sure! Do you want me to update the `getAuthUrl` method too?

atymic: Sure feel free while you're there :)

stidges:
Pushed a new commit that uses `sprintf`!

atymic:
Thanks!

PR #474, opened 2020-07-24
Always prefer stricter comparisons

lucasmichot:
Always prefer stricter comparisons

### DIFF
--- a/src/Gitee/Provider.php
+++ b/src/Gitee/Provider.php
@@ -50,7 +50,7 @@ class Provider extends AbstractProvider
 
         $user = json_decode($response->getBody(), true);
 
-        if (in_array('emails', $this->scopes)) {
+        if (in_array('emails', $this->scopes, true)) {
             $user['email'] = $this->getEmailByToken($token);
         }
 
@@ -78,7 +78,7 @@ class Provider extends AbstractProvider
         }
 
         foreach (json_decode($response->getBody(), true) as $email) {
-            if ($email['state'] === 'confirmed' && in_array('primary', $email['scope'])) {
+            if ($email['state'] === 'confirmed' && in_array('primary', $email['scope'], true)) {
                 return $email['email'];
             }
         }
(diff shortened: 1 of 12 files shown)


src/Gitee/Provider.php
```diff
@@ -78,7 +78,7 @@ protected function getEmailByToken($token)
         }
 
         foreach (json_decode($response->getBody(), true) as $email) {
-            if ($email['state'] == 'confirmed' && in_array('primary', $email['scope'])) {
+            if ($email['state'] === 'confirmed' && in_array('primary', $email['scope'])) {
```
atymic, line 81: Maybe worth adding `true` for strict in array comparison as well?

src/Jira/RsaSha1Signature.php
```diff
@@ -92,10 +92,10 @@ protected function baseString(Uri $url, $method = 'POST', array $parameters = []
     {
         $baseString = rawurlencode($method).'&';
         $schemeHostPath = $url->getScheme().'://'.$url->getHost();
-        if ($url->getPort() != '') {
```
atymic, line 95: I've not read the code, but is it possible this is a change in behaviour (ie nulls will now meet the condition)?

src/Gitee/Provider.php
lucasmichot: I don't think so, strict comparison with `===` is about the type, as `$strict` argument is about the case.

src/Jira/RsaSha1Signature.php
lucasmichot: I've commited  ae15ea9abae633895bd23f183223be9376a4d92f to ensure that we compare with `null` (`getPort()` return an integer or `null`)

src/Gitee/Provider.php
atymic: Sorry, not sure what you mean here? Below will use strict comparison instead of loose:

> If the third parameter strict is set to TRUE then the in_array() function will also check the types of the needle in the haystack.

```suggestion
            if ($email['state'] === 'confirmed' && in_array('primary', $email['scope'], true)) {
```

lucasmichot: Oh sorry, I didn't get you before either.
I've also added the strict comparison when using `is_array` function. See 2fb83e2909329e147f9ab0ca942b8b3751f10d48

PR #476, opened 2020-07-25
Remove useless imports

lucasmichot:
This PR removes useless imports.


atymic:
I don't think this is worth it's own PR.

lucasmichot:
@atymic I will add this to a more general clean-up PR.

PR #479, opened 2020-07-25
Refactor SocialiteProviders\Apple\Provider::mapUserToObject()

masticore252:
Replace the call to `filled()` with a simpler non-empty string check as it's only available in Laravel>=5.5 and throws a `BadMethodCallException` in Laravel 5.4

This is necessary because this library has a dependency to socialiteproviders/manager which allows installation in Laravel 5.4 in the latest version

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -195,8 +195,10 @@ class Provider extends AbstractProvider
      */
     protected function mapUserToObject(array $user)
     {
-        if (request()->filled('user')) {
-            $userRequest = json_decode(request('user'), true);
+        $value = trim((string) request('user'));
+
+        if ($value !== '') {
+            $userRequest = json_decode($value, true);
 
             if (array_key_exists('name', $userRequest)) {
                 $user['name'] = $userRequest['name'];


src/Apple/Provider.php
```diff
@@ -195,8 +195,10 @@ public function user()
      */
     protected function mapUserToObject(array $user)
     {
-        if (request()->filled('user')) {
-            $userRequest = json_decode(request('user'), true);
+        $value = trim((string) request('user'));
```
atymic, line 198:
```suggestion
        $value = trim((string) request('user'));
```

atymic, line 198: Could we do an early return here?

```php
if(empty(request('user')) {
    return null
}
```

masticore252: we can't because it would change the signature of the method, it should always return an instance of `Laravel\Socialite\Two\User`, as defined in `Laravel\Socialite\Two\AbstractProvider` where this method is inherited from

P.S.: in this case it returns an instance of `SocialiteProviders\Manager\OAuth2\User` which extends from `Laravel\Socialite\Two\User`

PR #485, opened 2020-07-29
Code cleanup

lucasmichot:
it is valuable to do a code clean up before the next big release.
This PR is the first of a small batch.

This code cleanup does the following:

- Remove useless imports.
- Modernize list syntax, and use single variables.
- Use camelCase names for variables.
- Remove unused parameters.
- Remove useless variables.
- Remove useless properties.

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -169,7 +169,7 @@ class Provider extends AbstractProvider
         );
 
         if ($this->usesState()) {
-            [$uuid, $state] = explode('.', $appleUserToken['nonce']);
+            $state = explode('.', $appleUserToken['nonce'])[1];
             if (md5($state) === $this->request->input('state')) {
                 $this->request->session()->put('state', md5($state));
                 $this->request->session()->put('state_verify', $state);
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -169,7 +169,7 @@ class Provider extends AbstractProvider
         );
 
         if ($this->usesState()) {
-            list($uuid, $state) = explode('.', $appleUserToken['nonce']);
+            [$uuid, $state] = explode('.', $appleUserToken['nonce']);
             if (md5($state) === $this->request->input('state')) {
                 $this->request->session()->put('state', md5($state));
                 $this->request->session()->put('state_verify', $state);
--- a/src/Instagram/Provider.php
+++ b/src/Instagram/Provider.php
@@ -115,8 +115,6 @@ class Provider extends AbstractProvider
             $sig .= "|$key=$val";
         }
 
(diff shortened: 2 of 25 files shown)


src/Instagram/Provider.php
```diff
@@ -114,8 +114,9 @@ protected function generateSignature($endpoint, array $params)
         foreach ($params as $key => $val) {
             $sig .= "|$key=$val";
         }
-        $signing_key = $this->clientSecret;
 
-        return hash_hmac('sha256', $sig, $signing_key, false);
+        $signingKey = $this->clientSecret;
+
+        return hash_hmac('sha256', $sig, $signingKey, false);
```
atymic, line 120: could we just put the client secret straight into the call?
```suggestion
        return hash_hmac('sha256', $sig, $this->clientSecret, false);
```

src/OAuthgen/Provider.php
```diff
@@ -22,8 +22,8 @@ class Provider extends AbstractProvider
     /**
      * BASE_URL.
      */
-    protected $oauth_url = 'https://auth.oauthgen.com';
-    protected $graph_url = 'https://graph.oauthgen.com/api/v1';
+    protected $oauthUrl = 'https://auth.oauthgen.com';
```
atymic, line 25: these could be protected consts instead of props i guess

src/Apple/Provider.php
```diff
@@ -169,7 +169,7 @@ public function user()
         );
 
         if ($this->usesState()) {
-            list($uuid, $state) = explode('.', $appleUserToken['nonce']);
+            $state = explode('.', $appleUserToken['nonce'])[1];
             if (md5($state) === $this->request->input('state')) {
```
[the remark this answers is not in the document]
lucasmichot: It was not done before but could be done in another PR. What do you think ?

There are LOTS of undefined indexes errors that need fixing in the codebase.

src/Instagram/Provider.php
lucasmichot: See 00f70ffa95a2fb1471d5129fc043595b11dfe524

src/Apple/Provider.php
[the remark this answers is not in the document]
atymic: Yep, no worries :)

src/OAuthgen/Provider.php
lucasmichot: See 010d222075c4451018beadc743c05c5b3bdf5680

lucasmichot: @atymic I think the PR was merged before this commit

atymic: oops 🙈 
Chuck in your next PR :)

Issue #488, opened 2020-07-31
Audit and deprecate broken/dead providers


atymic:
Archive & deprecate dead providers.

Pending actions

Abandon on packagist:

- [x] https://packagist.org/packages/socialiteproviders/everyplay
- [x] https://packagist.org/packages/socialiteproviders/mixer
- [x] https://packagist.org/packages/socialiteproviders/gamewisp
- [x] https://packagist.org/packages/socialiteproviders/dailymile
- [x] https://packagist.org/packages/socialiteproviders/rdio
- [x] https://packagist.org/packages/socialiteproviders/jawbone
- [x] https://packagist.org/packages/socialiteproviders/hitbox

lucasmichot:
Also abandon https://packagist.org/packages/socialiteproviders/rdio

atymic:
Hey @m1guelpf 

Could you mark the above packages as abandoned on packagist please?
I don't have the perms to do so.

m1guelpf:
@atymic Done!

atymic:
Thanks a ton @m1guelpf :)

atymic:
Also this provider if you have chance @m1guelpf 
https://packagist.org/packages/socialiteproviders/moikrug

PR #491, opened 2020-07-31
fix: remove dailymile provider

atymic:
Service shut down in 2019.
http://www.dailymile.com/


lucasmichot:
@atymic I've also spotted **Rdio** as a dead service

atymic:
Chuck a comment over here, or feel free to PR:
https://github.com/SocialiteProviders/Providers/issues/488

atymic:
Thanks :)

PR #494, opened 2020-07-31
Abandon rdio

lucasmichot:
Abandon rdio

See : https://github.com/SocialiteProviders/Providers/issues/488#issuecomment-667047244


atymic:
Thanks, archived the repo and will add to the abandon list :)

Issue #480, opened 2020-07-29
Release new version of all providers


atymic:
Release new major version dropping support for < `7.2`

- [x] Release Provider [23andme](https://github.com/SocialiteProviders/23andme/releases)
- [x] Release Provider [37Signals](https://github.com/SocialiteProviders/37Signals/releases)
- [x] Release Provider [500px](https://github.com/SocialiteProviders/500px/releases)
- [ ] Release Provider [Acclaim](https://github.com/SocialiteProviders/Acclaim/releases)
- [ ] Release Provider [Admitad](https://github.com/SocialiteProviders/Admitad/releases)
- [ ] Release Provider [Amazon](https://github.com/SocialiteProviders/Amazon/releases)
- [ ] Release Provider [AngelList](https://github.com/SocialiteProviders/AngelList/releases)
- [ ] Release Provider [App.net](https://github.com/SocialiteProviders/App.net/releases)
- [ ] Release Provider [Apple](https://github.com/SocialiteProviders/Apple/releases)
- [ ] Release Provider [ArcGIS](https://github.com/SocialiteProviders/ArcGIS/releases)
- [ ] Release Provider [Asana](https://github.com/SocialiteProviders/Asana/releases)
- [ ] Release Provider [Auth0](https://github.com/SocialiteProviders/Auth0/releases)
- [ ] Release Provider [Aweber](https://github.com/SocialiteProviders/Aweber/releases)
- [ ] Release Provider [Battle.net](https://github.com/SocialiteProviders/Battle.net/releases)
- [ ] Release Provider [Binance](https://github.com/SocialiteProviders/Binance/releases)
- [ ] Release Provider [Bit.ly](https://github.com/SocialiteProviders/Bit.ly/releases)
- [ ] Release Provider [Bitbucket](https://github.com/SocialiteProviders/Bitbucket/releases)
- [ ] Release Provider [Box](https://github.com/SocialiteProviders/Box/releases)
- [x] Release Provider [Buffer](https://github.com/SocialiteProviders/Buffer/releases)
- [ ] Release Provider [CampaignMonitor](https://github.com/SocialiteProviders/CampaignMonitor/releases)
- [ ] Release Provider [Cheddar](https://github.com/SocialiteProviders/Cheddar/releases)
- [ ] Release Provider [ClaveUnica](https://github.com/SocialiteProviders/ClaveUnica/releases)
[...]

PR #509, opened 2020-08-04
Fix POST key naming issue

lucasmichot:
As mentioned in https://github.com/SocialiteProviders/Providers/issues/480#issuecomment-668102861 - we should always check the version of Guzzle that is used, as the body can be passed either in `body`, either in `form_params`.

Not doing so would make some providers unusable with some versions of Guzzle.

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -3,6 +3,7 @@
 namespace SocialiteProviders\Apple;
 
 use Firebase\JWT\JWK;
+use GuzzleHttp\ClientInterface;
 use Illuminate\Http\Response;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Cache;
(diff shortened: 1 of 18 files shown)


atymic:
Is it possible you misread the guzzle upgrade notes? I checked, and it seems it changes from `body` to `form_params` in guzzle 6. Both guzzle 6 and 7 seems to use `form_params`?

See https://github.com/guzzle/guzzle/blob/master/docs/request-options.rst#form_params
and https://github.com/guzzle/guzzle/blob/master/UPGRADING.md

lucasmichot:
You are absolutely right @atymic - it seems that even more providers will need to be fixed :-( 
Steem, FaceIt, Fitbit, TeamWeek and Yiban.

atymic:
What do you mean? I don't think we need to change anything. Only guzzle 5 and under use `body`?

atymic:
I still don't think this is required @lucasmichot 
There is no changes in the key names between guzzle 6 and 7

PR #513, opened 2020-08-05
Add avatar (default or custom) and name if exists

J2TeamNNL:


src/Reddit/Provider.php
```diff
@@ -73,7 +73,7 @@ protected function mapUserToObject(array $user)
 
         return (new User())->setRaw($user)->map([
             'id'   => $user['id'], 'nickname' => $user['name'],
-            'name' => null, 'email' => null, 'avatar' => $avatar,
+            'name' => empty($user['subreddit']['title']) ? null : $user['subreddit']['title'], 'email' => null, 'avatar' => strtok($user['icon_img'], '?'),
```
atymic, line 76: What is the content of `$user['subreddit']['title']`? Doesn't look like name.

atymic, line 76: We're already including the avatar?

J2TeamNNL: Oh sorry about that, I didn't notice

J2TeamNNL: image
This is display name that user can custom

atymic: Are you sure that that ends up in the `subreddit` key?
That doesn't seem to make much sense.

J2TeamNNL: image
I pretty sure it is

src/Reddit/Provider.php
```diff
@@ -70,10 +69,16 @@ protected function mapUserToObject(array $user)
                 $avatar = substr($avatar, 0, $querypos);
             }
         }
+        
+        $name = null;
+        //Check if user have display name
```
atymic, line 74:
```suggestion
        //Check if user has a display name
```

PR #517, opened 2020-08-11
Detailed Telegram README

Max13:


src/Telegram/README.md
```diff
@@ -8,13 +8,20 @@ composer require socialiteproviders/telegram
...
+Then, you need to add your bot's configuration to `config/services.php`. The bot username is required, `client_id` must be `null`. The provider will also ask permission for the bot to write to the user.
 
 ```php
-'telegram' => [    
-  'bot' => env('TELEGRAM_BOT'),  
-  'client_secret' => env('TELEGRAM_CLIENT_SECRET'),  
-  'redirect' => env('TELEGRAM_REDIRECT_URI') 
+'telegram' => [
+    'bot' => env('TELEGRAM_BOT_NAME'),  // The bot's username
+    'client_id' => null,
```
atymic, line 22: guess we can remove this config option?

Max13: I think removing `client_id` would raise an exception because of https://github.com/SocialiteProviders/Manager/blob/master/src/ConfigTrait.php#L23

atymic: Ahh my bad, ignore :P

Max13: No problem 😁

PR #520, opened 2020-08-14
add okta custom auth server

kjiang14:
Adding support for Okta custom authorization server as described in this PR: https://github.com/SocialiteProviders/Okta/pull/2

### DIFF
--- a/src/Okta/Provider.php
+++ b/src/Okta/Provider.php
@@ -47,9 +47,17 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
+    protected function getAuthServerId()
+    {
+        return $this->getConfig('auth_server_id', 'default');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public static function additionalConfigKeys()
     {
-        return ['base_url'];
+        return ['base_url', 'auth_server_id'];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


atymic (commented):
Could you link to the docs covering this please?

atymic:
cc @kjiang14

robiiinos:
@atymic mmh, wait... this PR add custom auth server to Okta, but the updated README is for the YouTube Provider...

atymic:
🤦 

I'll fix that shortly haha. Good catch.

src/Okta/Provider.php
```diff
@@ -44,36 +44,44 @@ protected function getOktaUrl()
         return $this->getConfig('base_url');
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthServerId()
+    {
+        return $this->getConfig('auth_server_id', 'default');
```
gavinr, line 52: This should not default to `default` - per [this doc](https://support.okta.com/help/s/article/400-error-on-authentication-to-Okta-using-OIDC?language=en_US), many okta users should not have `default` for the `auth_server_id`. And you can't override this to an empty string in the config - if you send in an empty string in the config, `default` is set.

gavinr: Nevermind, looks like this was fixed in 8bdcb1ab129638daa11882168b80d3e4609d559b

PR #521, opened 2020-08-14
Add OVH (and subservices) provider

Max13:


src/Ovh/Provider.php
```diff
@@ -0,0 +1,119 @@
...
+            'endpoint',
+        ];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state)
+    {
+        $ovh = new Ovh(
```
atymic, line 34: Generally we don't use external libraries unless we can't do plain requests. Is there a reason why you used the OVH library?

src/Ovh/README.md
```diff
@@ -0,0 +1,65 @@
+# WHMCS
```
atymic, line 1:
```suggestion
# OVH
```

src/Ovh/Provider.php
Max13: It was to avoid reimplementing the checks and URLs. While it also uses Guzzle, it's more of a helper than anything else.
If you prefer, I can reimplement it in a class instead of an external library.

Max13: @atymic would it be acceptable to copy paste the Api.php file from OVH library? Or do you prefer I reimplement necessary methods only?

atymic: On seconds though, I think it's fine :)

Max13: Then my baby girl is ready to be merged 😂

PR #525, opened 2020-08-15
Typo in ALL providers

Max13:
I saw this typo in every providers README.

The driver is called this way: `Socialite::with('provider')` while [`with`](https://laravel.com/docs/socialite#optional-parameters) serves another purpose.

The correct way to call Socialite driver is: `Socialite::driver('provider')`

### DIFF
--- a/src/Acclaim/README.md
+++ b/src/Acclaim/README.md
@@ -38,7 +38,7 @@ protected $listen = [
 You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
 
 ```php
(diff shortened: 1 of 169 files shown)


Max13:
I'm on it

Max13:
@atymic Did I already PR'd a change for the docgen template or am I missing the concerned file?

atymic:
> @atymic Did I already PR'd a change for the docgen template or am I missing the concerned file?

`tools/docgen.php` :)

Max13:
Here you are ee881c2

PR #532, opened 2020-08-27
Added necessary step to Twitter docs

yoeriboven:
Twitter login won't work without this step and will return a vague error instead.


atymic (approved):
Thanks!

Issue #530, opened 2020-08-26
Apple Provider getting `OutOfBoundsException: Requested claim is not configured` when using userFromToken in 3.0.2


arun07as:
Hi

In the [apple provider](https://github.com/SocialiteProviders/Apple), in [yesterdays commit](https://github.com/SocialiteProviders/Apple/commit/fa7ac7e23390d7d81e647b09ec50f79a8de6b9ef) version v3.0.2, the `userFromToken` method keeps giving me the following exception

```
[2020-08-26 17:07:02] local.ERROR: OutOfBoundsException: Requested claim is not configured in C:\laragon\www\projectNameChanged:P\vendor\lcobucci\jwt\src\Token.php:176
Stack trace:
#0 C:\laragon\www\projectNameChanged:P\vendor\socialiteproviders\apple\Provider.php(144): Lcobucci\JWT\Token->getClaim('kid')
#1 C:\laragon\www\projectNameChanged:P\vendor\socialiteproviders\apple\Provider.php(110): SocialiteProviders\Apple\Provider::verify('eyJraWQiOiI4NkQ...')
#2 C:\laragon\www\projectNameChanged:P\vendor\laravel\socialite\src\Two\AbstractProvider.php(231): SocialiteProviders\Apple\Provider->getUserByToken('eyJraWQiOiI4NkQ...')
#3 C:\laragon\www\projectNameChanged:P\app\Http\Controllers\API\UserAPIController.php(803): Laravel\Socialite\Two\AbstractProvider->userFromToken('eyJraWQiOiI4NkQ...')
```

This was working fine on version v3.0.1. I tried downgrading to version v3.0.1 and it works fine now.

After looking into the code, the issue seems to be in [Provider.php line 144](https://github.com/SocialiteProviders/Apple/blob/fa7ac7e23390d7d81e647b09ec50f79a8de6b9ef/Provider.php#L144)

```
$kid = $token->getClaim('kid');
```

This tries to find the `kid` from inside the body of the token, but when I checked, it seems that the `kid` is present in the header.
![image](https://user-images.githubusercontent.com/17740186/91304529-40ce6e00-e7c7-11ea-825a-009fffe9977c.png)

So I'm not exactly sure, but changing line 144 to

```
$kid = $token->getHeader('kid');
```

might just fix the issue

atymic:
Could you please create a PR? Thanks!

atymic:
cc @pyrou

pyrou:
[...]

PR #534, opened 2020-08-29
Fix JWT kid reading

pyrou:
This pull request fixes https://github.com/SocialiteProviders/Providers/issues/530 introduced by PR https://github.com/SocialiteProviders/Providers/pull/528. 

The `kid` claim (key Id) is always provided in header section of a JWT token, and wrong assumption was made about ability of retrieving this value using the `getClaim()`method of lcobucci/jwt library.

As guessed by @arun07as, using [getHeader()](https://github.com/lcobucci/jwt/blob/3.3.3/src/Token.php#L106) method resolve the issue.

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -141,7 +141,7 @@ class Provider extends AbstractProvider
 
         $publicKeys = JWK::parseKeySet($data);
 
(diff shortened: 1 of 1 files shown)


atymic (approved):
Thanks!

PR #542, opened 2020-09-18
fix issue: array_key_exists null given

orhan-swe:
$value is not a json string but a apple user id string and in that case json_decode will return null, so a check is needed

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -198,7 +198,7 @@ class Provider extends AbstractProvider
         if ($value !== '') {
             $userRequest = json_decode($value, true);
 
-            if (array_key_exists('name', $userRequest)) {
+            if ($userRequest && array_key_exists('name', $userRequest)) {
                 $user['name'] = $userRequest['name'];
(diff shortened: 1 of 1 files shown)


src/Apple/Provider.php
```diff
@@ -198,7 +198,7 @@ protected function mapUserToObject(array $user)
         if ($value !== '') {
             $userRequest = json_decode($value, true);
 
-            if (array_key_exists('name', $userRequest)) {
+            if ($userRequest && array_key_exists('name', $userRequest)) {
```
atymic, line 201: could we instead do `isset($userRequest['name'])` to cover both those cases?

orhan-swe:
$userRequest is null so that may throw a null exception, I am not sure need to test it..

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

atymic:
> $userRequest is null so that may throw a null exception, I am not sure need to test it..

isset will return false on a null return

atymic (approved):

PR #556, opened 2020-10-17
fix(telegram): error when missing field

### LINKED ISSUE #555: A bug fix is needed for Telegram provider
The Telegram provider has a bug that prevents users who do not have a last name from logging in. If it is not there, then the script crashes with an error. I forked the repository for myself, but I would like the bug to be fixed in the official version. Is it possible?

Here's the fix https://github.com/Altair4713/Telegram/commit/8755269312d91534de6058323d6fdebf0faf97f8#diff-3c2c7bc87589e022ac1cdaf01268b250cd8456756927542cb2c58c67e67cf33f

robiiinos:
Fix #555

### DIFF
--- a/src/Telegram/Provider.php
+++ b/src/Telegram/Provider.php
@@ -83,10 +83,12 @@ class Provider extends AbstractProvider
      */
     protected function mapUserToObject(array $user)
     {
+        $name = trim(sprintf('%s %s', $user['first_name'] ?? '', $user['last_name'] ?? ''));
+
         return (new User())->setRaw($user)->map([
             'id'        => $user['id'],
             'nickname'  => $user['username'],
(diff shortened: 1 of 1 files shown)


src/Telegram/Provider.php
```diff
@@ -83,10 +83,12 @@ public function redirect()
      */
     protected function mapUserToObject(array $user)
     {
+        $name = trim(sprintf('%s %s', $user['first_name'] ?? '', $user['last_name'] ?? ''));
+
         return (new User())->setRaw($user)->map([
             'id'        => $user['id'],
             'nickname'  => $user['username'],
-            'name'      => $user['first_name'].' '.$user['last_name'],
+            'name'      => $name ?? null,
```
atymic, line 91: name will always be set here, so we need an empty check :)

PR #557, opened 2020-10-18
feat(streamlabs): add third-party accounts

robiiinos:
This PR allow developers to access the third-party accounts that the user signed-in with Streamlabs, so they can access more information (such as nicknames, profile picture, ...). Documentation has been updated as well.

Documentation: https://dev.streamlabs.com/reference#user


atymic (approved):
Thanks!

PR #558, opened 2020-10-18
feat(streamelements): add user fields

robiiinos:
Added fields to user:
- ``alias``
- ``type``
- ``verified``
- ``partner``
- ``suspended``

Documentation: https://docs.streamelements.com/reference/channels#get-user-channel-details


atymic (approved):
Thanks!

Issue #562, opened 2020-10-21
Steam Provider - wrong config key and problem with userFromToken()


michaloravec:
There are 2 main problems in that provider.

Instead of `api_key`  it should be `client_secret` as a key in config. Look [here](https://github.com/SocialiteProviders/Providers/issues/527). She made PR for that, but it was closed.

https://github.com/SocialiteProviders/Steam/pull/4

Also there is problem with `userFromToken($token)` it doesn't work because inside that method a variable `$token` in not used, you can see it [here](https://github.com/SocialiteProviders/Steam/blob/master/Provider.php#L106).

atymic:
The PR was on the subtree repo (readonly). Please re-open here and I will review and merge :)

michaloravec:
Steam on master branch doesn't work at all!

michaloravec:
@atymic How I said you have to merge it to the version 3.0.2. Because last working version was 3.0.2. But you merged it to the master branch.

Take a look https://github.com/SocialiteProviders/Steam/compare/v3.0.2...v3.1.0

There are other differences and it causes that it doesn't work at all right now.

atymic:
@michaloravec you branched from the master branch, so that was where it was applied.

Feel free to make the other changes required to fix issues in the lastest, because we need to fix it there.

michaloravec:
@atymic I'm not sure but maybe OpenID 2.0 supports only https.

Yeah because there wasn't other option then branch from master branch.

michaloravec:
@atymic It was tested on live server and it doesn't work when there are changes from https://github.com/SocialiteProviders/Steam/commit/351736c7ef20ca399c6dd55c18c3b3c1fb9758ac

So I'm going to make PR for that.

atymic:
Let me know if it's all good now and i'll close this @michaloravec :)

michaloravec:
@atymic Yes, everything is good now. Thanks.

atymic:
Thanks for your help :)

EdgarSedov:
This PR had broken steam auth on my site (which obviously used api_key as config parameter)
Please do such changes with major version, as this is BC and can broke consumers apps

PR #563, opened 2020-10-22
Steam Provider - fix config key and problem with userFromToken()

michaloravec:
It needs to be merged to the latest release version which is 3.0.2

More info is in this issue #562


atymic (approved):
Thanks for the PR!

PR #566, opened 2020-10-22
ref(naver): refactor oauth2 workflow

robiiinos:
Refactor Naver OAuth2 provider, as they are now providing a JSON api (only XML previously).

This provider can now be generic like others.
- Dropped XML for JSON (+ deps).
- Updated OAuth2 URL's and workflow.
- Updated README with link to documentation.

OAuth2 API endpoints documentation: https://developers.naver.com/docs/login/api/
User endpoint documentation (part 5.): https://developers.naver.com/docs/login/profile/

> Note: Followed the same guidelines as the other providers that I am maintaining.


atymic (approved):
Great job. You're on 🔥 
Thanks!

PR #567, opened 2020-10-23
Steam Provider - always use http #562

michaloravec:
More info is in this issue #562

### DIFF
--- a/src/Steam/Provider.php
+++ b/src/Steam/Provider.php
@@ -38,7 +38,7 @@ class Provider extends AbstractProvider
     /**
      * @var string
      */
-    public const STEAM_INFO_URL = 'https://api.steampowered.com/ISteamUser/GetPlayerSummaries/v0002/?key=%s&steamids=%s';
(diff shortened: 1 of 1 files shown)


FoksVHox (approved):
Seems fine to me

atymic:
Can you check what redirect URLs your server is generating?
I checked and HTTPs is supported by them (and much more secure). There must be a different issue in our implementation

michaloravec:
@atymic This provider is based on this package https://github.com/invisnik/laravel-steam-auth

And they also use `http`, except **OPENID_URL** which is through `https` https://github.com/invisnik/laravel-steam-auth/blob/master/src/SteamAuth.php#L49

And I don't change **OPENID_URL** in this pull request https://github.com/SocialiteProviders/Providers/blob/master/src/Steam/Provider.php#L36 it's still on `https`

atymic (approved):

atymic:
Thanks!

PR #539, opened 2020-09-08
feat: upgrade manager version to 4.0

atymic:

PR #522, opened 2020-08-15
Added Binance Provider

robiiinos:
Added Binance as a Provider. 

SocialiteProviders documentation PR: https://github.com/SocialiteProviders/website/pull/4

PR #568, opened 2020-10-23
fix(composer): drop old manager version

robiiinos:
First draft at trying to normalize all providers; unsure about compatibility problems so please let me know if any questions.

Comment: https://github.com/SocialiteProviders/Providers/pull/522#discussion_r470958724

This is while we are waiting for an official release of 4.0, https://github.com/SocialiteProviders/Providers/pull/539

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "require": {
         "php": "^7.2",
         "ext-json": "*",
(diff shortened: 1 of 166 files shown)


atymic:
Closed in #539 but thanks!

PR #569, opened 2020-10-23
fix(subtree): providers list

robiiinos:
2 commits so this is easier for you to review;

1. Added missing provider to monorepo-builder (Amazon / Orcid)
2. Ordered list alphabetically


atymic (approved):
Thanks! I added repos for the two that got missed.

PR #581, opened 2020-11-05
Use Zoho-provided accounts server when provided

coclav-skore:
In some instance, Zoho will return which accounts server to use to get further user's details.

For example, our customer in Europe is instructed to use https://accounts.zoho.eu (instead of .com) 

This change is about using the `accounts-server` provided by Zoho, if any. Otherwise will use the same as today by default.

Note, in my tests, the initial authUrl `getAuthUrl()` does not need to be taylored to the user's instance.

Thanks for massive amount of work done on Socialite!

Best Regards, 

Colin

### DIFF
--- a/src/Zoho/Provider.php
+++ b/src/Zoho/Provider.php
@@ -24,13 +24,21 @@ class Provider extends AbstractProvider
     {
         return $this->buildAuthUrlFromBase('https://accounts.zoho.com/oauth/v2/auth', $state);
     }
+    
+    /**
+     * Gets the Accounts Server to use from Zoho provider
+     */
+    protected function getAccountsServerUrl()
+    {
+        return $this->request->input('accounts-server', 'https://accounts.zoho.com');
+    }
 
     /**
      * {@inheritdoc}
      */
     protected function getTokenUrl()
     {
-        return 'https://accounts.zoho.com/oauth/v2/token';
+        return $this->getAccountsServerUrl() . '/oauth/v2/token';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


coclav-skore:
(sorry i don't know how to fix the last styling issue)

src/Zoho/Provider.php
```diff
@@ -24,21 +24,29 @@ protected function getAuthUrl($state)
     {
         return $this->buildAuthUrlFromBase('https://accounts.zoho.com/oauth/v2/auth', $state);
     }
+
```
atymic, line 27:
```suggestion

```

atymic (approved):
Looks like a whitespace issue 👍

src/Zoho/Provider.php
```diff
@@ -24,21 +24,28 @@ protected function getAuthUrl($state)
     {
         return $this->buildAuthUrlFromBase('https://accounts.zoho.com/oauth/v2/auth', $state);
     }
+    /**
```
atymic, line 27:
```suggestion

    /**
```

atymic:
Having a look at the docs, I don't see this param (just a `location` one)?
https://www.zoho.com/accounts/protocol/oauth/multi-dc/client-authorization.html

coclav-skore:
I wish it was documented and i didn't spend hours trying to figure it out...

This is the dump of the Socialite driver callback

```
$p = Socialite::driver('zoho');
dump($p);
```

Screenshot 2020-11-05 at 16 00 22

atymic:
> I wish it was documented and i didn't spend hours trying to figure it out...

No worries, I can't see any harm in merging because even if they remove it it will just use the default 👍 
Thanks!

Issue #589, opened 2020-11-30
Mixcloud provider is broken


mattsims:
The Mixcloud provider is broken, as it's attempting to use a deprecated method to obtain the access token – `Laravel\Socialite\Two\AbstractProvider::getAccessToken()` – which was removed in [this commit](https://github.com/laravel/socialite/commit/0c234df124df24eed19355949d412ce66cee26fc#diff-8d3fd95aa368bbc8b78e7755afdf6132dbab48487f3f53591e55bf3f757f7121).

Ive submitted [this PR](https://github.com/SocialiteProviders/Mixcloud/pull/2) with updated code to use the latest method(s).

atymic:
Please PR to this repo, you PR'd to the read only fork :)

mattsims:
D'oh. New PR #590 created.

PR #590, opened 2020-12-01
Fix parsing of Mixcloud access token

mattsims:
The Mixcloud provider was broken, as it was attempting to use a deprecated method to obtain the access token – `Laravel\Socialite\Two\AbstractProvider::getAccessToken()` – which was removed in [this commit](https://github.com/laravel/socialite/commit/0c234df124df24eed19355949d412ce66cee26fc#diff-8d3fd95aa368bbc8b78e7755afdf6132dbab48487f3f53591e55bf3f757f7121).

This PR updates the code to use the latest method(s), resolving issue #589.


atymic (approved):
Thanks!

PR #592, opened 2020-12-02
Update dependency of lcobucci/jwt lib on Apple Provider

DoctrineDark:


atymic:
Is it worth bumping to v4 for php 8 support here?

DoctrineDark:
> Is it worth bumping to v4 for php 8 support here?

I think it should be a separate pr, because this changes fix compatibility with lcobucci/jwt. Without them lib does not work correctly.

atymic:
Ahh, so it's currently broken?
As in, they had a breaking change in a minor version? Or there are bugs in our code?

DoctrineDark:
> Ahh, so it's currently broken?
> As in, they had a breaking change in a minor version? Or there are bugs in our code?

@atymic , There are no bugs, but passport and SocialiteProviders/Apple both installed requires different version of lcobucci/jwt. Now there is this error: https://github.com/lcobucci/jwt/blob/3.4/src/Token.php#L295, and thats why this changes need to be approved.

atymic (approved):

PR #596, opened 2020-12-08
add phone to return fields

sorasu:
to the developers whose app got approved, Line grants permission to get user's phone number (OC_PHONE_NUMBER)


atymic:
@sorasu I would avoid adding to the default fields unless it's there for everyone. You should be able to add the additional scope and access it directly?

sorasu:
u r correct. sorry I didn't know that I could just add additional scopes directly.
appreciate for your prompt response.

PR #597, opened 2020-12-08
add phone to return fields

sorasu:
to the developers whose app got approved, Line grants permission to get user's phone number (OC_PHONE_NUMBER)


sorasu:
sorry I didn't know that I could just add additional scopes directly.
no need to edit the provider

PR #598, opened 2020-12-11
Add Fablabs.io provider

MacTwister:
Adds support for using [fablabs.io](https://fablabs.io/about) as a login provider

### DIFF
--- a/monorepo-builder.yml
+++ b/monorepo-builder.yml
@@ -38,6 +38,7 @@ parameters:
     src/Etsy: 'git@github.com:SocialiteProviders/Etsy.git'
     src/Eventbrite: 'git@github.com:SocialiteProviders/Eventbrite.git'
     src/EyeEm: 'git@github.com:SocialiteProviders/EyeEm.git'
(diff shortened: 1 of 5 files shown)


src/Fablabsio/FablabsioExtendSocialite.php
```diff
@@ -0,0 +1,18 @@
...
+class FablabsioExtendSocialite
+{
+    /**
+     * Register the provider.
+     *
+     * @param \SocialiteProviders\Manager\SocialiteWasCalled $socialiteWasCalled
+     */
+    public function handle(SocialiteWasCalled $socialiteWasCalled)
+    {
+        $socialiteWasCalled->extendSocialite('fablabsio', Provider::class);
```
atymic, line 16: Probably could just remove the `io` from the provider folder and here?
I don't think it conflicts with any other provider?

atymic (approved):

MacTwister:
Thank you, and for your time ✌️

PR #605, opened 2020-12-27
Adds zoom provider

craigpaul:


src/Zoom/Provider.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+namespace SocialiteProviders\Zoom;
+
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
```
atymic, line 8: Could we have an `@see` tag with a link to their docs please 👍

src/Zoom/README.md
```diff
@@ -0,0 +1,42 @@
...
+    ],
+];
+```
+
+### Usage
+
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
+
+```php
+return Socialite::driver('zoom')->redirect();
```
atymic, line 41: Could you add the returned fields here?
Have a look at some other providers for examples :)

atymic (approved):
Just a few minor bits :)

PR #609, opened 2021-01-06
Adds null fallback for avatar as zoom does not send it if it does not exist

craigpaul:
Turns out Zoom does not send the `pic_url` field if you do not have one set in their platform 🤷


atymic (approved):
Thanks!

PR #610, opened 2021-01-08
Added provider for ProSanteConnect. https://www.ingroupe.com/produits/pro-sante-connect

cleitus:


cleitus:
please check for a quick introduction: https://integrateurs-cps.asipsante.fr/pages/prosanteconnect/presentation-generale

.gitignore
```diff
@@ -1,2 +1,3 @@
 /vendor
 composer.lock
+.idea
```
oliworx, line 3: this should be part of your own local gitignore, since it has nothing to do with the project, but with your local development environment

atymic:
Can you explain what this is please?

cleitus:
> Can you explain what this is please?

Hello, 

Simply put ProSanteConnect is the equvalent of FranceConnect but for all healthcare workers in France. With Covid, this system was put in place by the french government in order as an alternative form of authentification via e-cps (Electronic CPS - compared to the more common  physical card : CPS). Since then it has seen exponential use and we now serve over a million health professionnals each month and growing. With this eco system now in place we need all healthcare services to be able to implement authentication via e-cps as quickly as possible, thus why I added this provider.

PR #612, opened 2021-01-09
Added Saml2 provider

27pchrisl:
This is not an OAuth provider, but it does meet the Socialite specification in terms of implementing the contract by extending `Laravel\Socialite\Two\AbstractProvider`, implementing `Laravel\Socialite\Contracts\Provider` and following the same configuration method.

For testing you should be able to use any SAML provider (Google, Azure), but there is a test service that works here too: https://samltest.id

https://samltest.id is a good example of a service that doesn't implement the HTTP-Redirect part of the specification, as noted in the "SAML Protocol" part of the README.md, but Google, Azure and many others do.

For testing this with samltest.id:
* Install the provider, and configure according to the docs using `https://samltest.id/saml/idp` as the `metadata` value
* Expose the Laravel app using ngrok or similar, setting `$proxies = '*'` in `TrustProxies.php` for the purposes of testing.
* Add the metadata route in Laravel (see docs)
* Change the callback route to use POST (required by samltest.id), and add the callback to the Csrf exception list in `VerifyCsrfToken.php` if you are using the web.php routes file
* Submit the metadata URL (https://xyz.ngrok.io/auth/saml2/metadata) to https://samltest.id/upload.php
* Access https://xyz.ngrok.io/auth/redirect in your browser to start the flow.

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -11,7 +11,9 @@ use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Facades\URL;
 use Illuminate\Support\Facades\Validator;
+use Illuminate\Support\Str;
 use Laravel\Socialite\Contracts\Provider as SocialiteProvider;
+use Laravel\Socialite\Contracts\User as SocialiteUser;
 use Laravel\Socialite\Two\AbstractProvider;
 use Laravel\Socialite\Two\InvalidStateException;
 use LightSaml\Binding\BindingFactory;
@@ -58,6 +60,9 @@ class Provider extends AbstractProvider implements SocialiteProvider
      * @var array
      */
     protected $config;
+
+    const CACHE_KEY = 'socialite_saml2_metadata';
+    const CACHE_KEY_TTL = self::CACHE_KEY . '_ttl';
 
     public function __construct(Request $request)
     {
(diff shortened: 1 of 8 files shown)


atymic:
Hey @27pchrisl 

Sorry, I've been busy as of a late. Will review soon :)

27pchrisl:
No worries @atymic !

src/Saml2/Provider.php
```diff
@@ -0,0 +1,389 @@
...
+        $entityId = $this->getConfig('entityid');
+        $certificate = $this->getConfig('certificate');
+
+        if (!$entityId || !$certificate) {
+            throw new MissingConfigException('When using "acs", both "entityid" and "certificate" must be set');
+        }
+
+        if (!Str::startsWith($certificate, '-----BEGIN CERTIFICATE-----')) {
+            $certificate = "-----BEGIN CERTIFICATE-----\r\n".
+                implode("\r\n", str_split($certificate, 80)).
```
atymic, line 145: Could we have a comment on the why this magic is done please? :)

src/Saml2/Provider.php
```diff
@@ -0,0 +1,389 @@
...
+        $xml = Cache::get(self::CACHE_KEY);
+        $ttl = Cache::get(self::CACHE_KEY_TTL);
+
+        $deserializationContext = new DeserializationContext();
+
+        if ($xml && $ttl && $ttl + $this->getConfig('ttl', 86400) > time()) {
+            return Metadata::fromXML($xml, $deserializationContext);
+        }
+
+        Cache::forever(self::CACHE_KEY_TTL, time());
```
atymic, line 174: Couldn't we use an expiring key here, rather than the forever cache?

atymic (approved):
Great work, looks good!

Just a few minor bits :)

src/Saml2/Provider.php
27pchrisl: The reason for this is that if your metadata cache is cleared, and due to some network or other issue you cannot refresh the metadata right now then you cannot serve any more authentication requests until you get the metadata again.

There may be a more Laravel idiomatic way of saying "refresh this cached item every X amount of time, but if you fail to refresh it then don't remove it" that I'm not aware of?

src/Saml2/Provider.php
27pchrisl: Yup, and I actually realised the library can provide this magic so I used the library method and commented as well

atymic (approved):

atymic:
https://github.com/SocialiteProviders/Saml2 
Live

PR #614, opened 2021-01-11
fix: incomplete telegram user data

DengSihan:
when a telegram user doesn't have `username` or `photo_url`, replace them with `first_name` and `null`


atymic (approved):
Thanks!

PR #618, opened 2021-01-17
Add Basecamp provider

mahmudz:


atymic:
Thanks, but this already exists:
https://github.com/SocialiteProviders/Providers/tree/master/src/ThirtySevenSignals

PR #623, opened 2021-02-01
Create Atlassian provider

ryancooley:
Differs from existing Jira provider in that it uses OAuth 2 as documented at https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/

### DIFF
--- a/src/Atlassian/README.md
+++ b/src/Atlassian/README.md
@@ -1,0 +1,42 @@
+# Atlassian
+
+```bash
+composer require socialiteproviders/atlassian
+```
+
+## Installation & Basic Usage
+
+Please see the [Base Installation Guide](https://socialiteproviders.com/usage/), then follow the provider specific instructions below.
+
+### Add configuration to `config/services.php`
+
+```php
+'atlassian' => [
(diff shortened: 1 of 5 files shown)


src/Atlassian/README.md
```diff
@@ -0,0 +1,42 @@
...
+];
+```
+
+### Usage
+
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
+
+```php
+return Socialite::driver('atlassian')->redirect();
+```
```
atymic, line 42: Could you add the returned fields here?

atymic (approved):
Thanks! Would it be worth deprecating the existing jira provider, or is that still useful?

atymic:
cc @ryancooley

atymic:
cc @ryancooley

atymic:
cc @ryancooley

src/Atlassian/README.md
atymic, line 42:
```suggestion
### Returned User fields
- ``id``
- ``nickname``
- ``name``
- ``email``
- ``picture``
```

atymic (approved):

PR #626, opened 2021-02-03
Update Google provider to bring it up-to-date with original Socialite Google provider

bmstanley:
The Google provider is now more in-line with the original Socialite Google provider. The token URL is now the v4 URL and the "nickname" value is now pulled from the "nickname" user data key.

### DIFF
--- a/src/Google/Provider.php
+++ b/src/Google/Provider.php
@@ -32,10 +32,7 @@ class Provider extends AbstractProvider
      */
     protected function getAuthUrl($state)
     {
-        return $this->buildAuthUrlFromBase(
-            'https://accounts.google.com/o/oauth2/auth',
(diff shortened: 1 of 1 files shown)


atymic (commented):
Seems they are using v3, is there any breaking changes between 3 and 4?
https://github.com/laravel/socialite/blob/master/src/Two/GoogleProvider.php

atymic:
cc @bmstanley

atymic:
cc @bmstanley

bmstanley:
Sorry for the delay... I am not 100% sure to be honest. I am using the upgraded version in my projects and have not had any issues thus far.

atymic:
Closed in #668 but thanks :)

PR #628, opened 2021-02-09
Adobe Provider

zeo:
This PR merges a brand new provider, Adobe.

### DIFF
--- a/src/Adobe/Provider.php
+++ b/src/Adobe/Provider.php
@@ -16,19 +16,19 @@ class Provider extends AbstractProvider implements ProviderInterface
 
     protected function getAuthUrl($state): string
     {
-        return $this->buildAuthUrlFromBase(self::BASE_URL . '/authorize', $state);
+        return $this->buildAuthUrlFromBase(self::BASE_URL.'/authorize', $state);
     }
 
     protected function getTokenUrl(): string
     {
-        return self::BASE_URL . '/token';
+        return self::BASE_URL.'/token';
     }
 
     protected function getTokenFields($code): array
     {
         return [
             ...parent::getTokenFields(),
-            'grant_type' => 'authorization_code'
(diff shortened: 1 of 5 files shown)


src/Adobe/Provider.php
```diff
@@ -0,0 +1,68 @@
...
+
+    protected function getTokenUrl(): string
+    {
+        return self::BASE_URL.'/token';
+    }
+
+    protected function getTokenFields($code): array
+    {
+        return [
+            ...parent::getTokenFields(),
```
atymic, line 30: Is this supported in 7.4? Just checking :P

src/Adobe/Provider.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace SocialiteProviders\Adobe;
+
+use GuzzleHttp\Exception\GuzzleException;
+use SocialiteProviders\Manager\Contracts\OAuth2\ProviderInterface;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider implements ProviderInterface
```
atymic, line 10: Can we have an `@see` with a link to the docs please

src/Adobe/composer.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "socialiteproviders/adobe",
```
atymic, line 2: Can you please add this to the split script (monorepo builder file in repo root)

atymic (changes requested):

src/Adobe/Provider.php
zeo: Just checked it out, seems to be supported since PHP 8.0. Will refactor it to support 7.4 syntax!

atymic (approved):

Issue #322, opened 2019-07-07
Nextcloud Oauth Provider


FrouxBY:
Hi, 

Does someone will have the time to add a Nextcloud provider to Socialite ?
I'll find it very helpful for use in some app like BookStack,

Thanks,

PR #632, opened 2021-02-26
Nextcloud OAuth2

artur-shaik:
related to #322

### DIFF
--- a/src/Nextcloud/Provider.php
+++ b/src/Nextcloud/Provider.php
@@ -56,7 +56,6 @@ class Provider extends AbstractProvider
      */
     protected function mapUserToObject(array $user)
     {
-        print_r($user);
         return (new User())->setRaw($user)->map([
             'id'       => $user['ocs']['data']['id'],
             'nickname' => $user['ocs']['data']['id'],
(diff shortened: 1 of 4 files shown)


src/Nextcloud/Provider.php
```diff
@@ -0,0 +1,89 @@
...
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state)
+    {
+        return $this->buildAuthUrlFromBase($this->getInstanceUri().'/index.php/apps/oauth2/authorize', $state);
```
atymic, line 29: Looking at the docs, you shouldn't need the `index.php`?
https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/oauth2.html

src/Nextcloud/Provider.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+namespace SocialiteProviders\Nextcloud;
+
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
```
atymic, line 8: Could you add an `@see` docblock with a link to the docs?

src/Nextcloud/Provider.php
artur-shaik: You are right, this is optional. Will fix it later, thanks.

atymic (approved):
Looks good :)

PR #638, opened 2021-03-12
Microsoft Ads is now compatible with Microsoft provider.

### LINKED ISSUE #637: Microsoft Provider missing scope in Guzzle form_params
Hey there,

I'm currently integrating OAuth2 for Microsoft Ads, and I've been getting exceptions like this:
```text
GuzzleHttp\Exception\ClientException
Client error: `POST https://login.microsoftonline.com/common/oauth2/v2.0/token` resulted in a `400 Bad Request` response: {"error":"invalid_scope","error_description":"AADSTS70011: The provided request must include a 'scope' input parameter.\ (truncated...)
```

The required scopes are:
```php
'scopes' => [
            'openid',
            'profile',
            'https://ads.microsoft.com/ads.manage',
            'offline_access',
        ],
```

Took me a while until I figured out that the `POST` request issued in the `getAccessTokenResponse()` method, does not include the `scope` inside `'form_params =>'.
https://github.com/SocialiteProviders/Providers/blob/246f67f1f25f6536910a990ff3c674794e00b428/src/Microsoft/Provider.php#L99

Adding the scope to the overriden method in `\SocialiteProviders\Microsoft\Provider` fixes this:
```php

    /**
     * {@inheritdoc}
     */
    protected function getTokenFields($code)
    {
        return array_merge(parent::getTokenFields($code), [
            'grant_type' => 'authorization_code',
            'scope' =>  $this->formatScopes($this->getScopes(), $this->scopeSeparator),
        ]);
    }

```

I'm open to make a pull-request for this, Let me know please :)

madsem:
Added support for Microsoft Ads, which requires `scope` to be included in the `POST` request to the token url.
Fixes #637

See the example:
https://docs.microsoft.com/en-us/advertising/guides/get-started?view=bingads-13#quick-start-production

```powershell
# Get the initial access and refresh tokens. 

$response = Invoke-WebRequest https://login.microsoftonline.com/common/oauth2/v2.0/token -ContentType application/x-www-form-urlencoded -Method POST -Body "client_id=$clientId&scope=https://ads.microsoft.com/ads.manage%20offline_access&code=$code&grant_type=authorization_code&redirect_uri=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fnativeclient"
```

PS: sorry for the PR inside Microsoft repository, seems like I did that wrong :)

### DIFF
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -97,6 +97,7 @@ class Provider extends AbstractProvider
     {
         return array_merge(parent::getTokenFields($code), [
             'grant_type' => 'authorization_code',
+            'scope' => parent::formatScopes(parent::getScopes(), $this->scopeSeparator),
         ]);
     }
 
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -97,7 +97,7 @@ class Provider extends AbstractProvider
     {
         return array_merge(parent::getTokenFields($code), [
             'grant_type' => 'authorization_code',
-            'scope' => parent::formatScopes(parent::getScopes(), $this->scopeSeparator),
+            'scope'      => parent::formatScopes(parent::getScopes(), $this->scopeSeparator),
         ]);
     }
 


atymic (approved):

Jamesking56:
This update broke Microsoft login on my clients site for certain Office 365 users so I had to downgrade to 4.1.0

No idea why it stops certain users from being able to login.

@atymic You may want to revert this if other users have the same problem as me

Jamesking56:
The error code some users got from Microsoft was `AADSTS65001` but after downgrading to 4.1.0 the error went away and they could login as normal

madsem:
Hey @Jamesking56 

This could be the last post here:
https://social.msdn.microsoft.com/Forums/sqlserver/en-US/2c051823-aac5-4a57-b0ea-8da305b663b1/invalind-grant-error-code-aadsts65001?forum=WindowsAzureAD

Would explain why only some of your users have this issue

Jamesking56:
@madsem I believe its only a single-tenant O365?

madsem:
@Jamesking56 For the OT yes, but I was talking about the last post in that thread:

> There is a case when your Web Application (A) and Web Api Application(B) are in different AAD tenant, then you must re-consent on B's tenant.
That's an headache for multi-tenant case...

Because as you write, only some of your users experience this issue. So I assume it has something to do with how your tenants gave consent. Otherwise it would just break for all of them.

Jamesking56:
@madsem I don't think I quite follow, we are only authenticating in Laravel so surely there is only 1 application involved? Also, why does 4.1.0 work as normal?

madsem:
@Jamesking56 as to why it happens. I honestly don't know, I think it might be that somehow the scope is now checked whereas it wasn't before (on Microsoft's end) because the scopes weren't included. I think there must be several factors at play here, and most likely I believe this is at Microsoft's end... If you check here:

https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow

The 'scope' is also listed as required for requesting an access token there... So I don't believe there's a problem with this PR, but rather that maybe wome of your tenants were giving consent maybe through a different Azure app, or something else was different. And now Microsoft's servers reject the access token.

Atm I think there is something going wrong at their end, and you have these issues with just a few of your tenants because they somehow authorized differently than the other tenants where it still works also with this PR. Does that make any sense to you?

Issue #640, opened 2021-03-18
HubSpot Compatibilty problem with joelbutcher/socialstream


nasatome:
on a clean install of https://github.com/joelbutcher/socialstream 3.0 with laravel/jetstream in Laravel v8.33.1 (PHP v7.4.15)

 the name must not be null and the id is required for getId function.
 `$account = Socialstream::findConnectedAccountForProviderAndId($provider, $providerAccount->getId());` 

fix in the line 58 of Provider.php:  https://github.com/SocialiteProviders/HubSpot
https://github.com/SocialiteProviders/HubSpot/blob/6e2ec0571af3a526e60e2bfbdfbe35f4ba44dead/Provider.php#L58
```php
return (new User())->setRaw($user)->map([
            'nickname' => null, 
            'name' => "",
            'email'    => $user['user'], 
            'avatar' => null,
            'id' => $user['user_id']
        ]);
```

temporal fix for me: 

I extended the class `JoelButcher\Socialstream\Http\Controllers\OAuthController` and substituted handleProviderCallback to add the following code
```php
            if(empty($providerAccount->id) && !empty($providerAccount->user["user_id"])){
                $providerAccount->id = $providerAccount->user["user_id"];
            }

            if(empty($providerAccount->name)){
                $providerAccount->name = "";
            }
```

in case it is useful to anyone

```php
<?php

namespace App\Http\Controllers\Hubspot;

use Illuminate\Http\Request;
use JoelButcher\Socialstream\Contracts\ResolvesSocialiteUsers;
use JoelButcher\Socialstream\Features;
use JoelButcher\Socialstream\Http\Controllers\OAuthController;
use JoelButcher\Socialstream\Socialstream;
use Laravel\Jetstream\Jetstream;
use Laravel\Socialite\Facades\Socialite;
use Laravel\Socialite\Two\InvalidStateException;
use Illuminate\Support\Facades\Auth;

class ExtendedOAuthController extends OAuthController
{

    /**
     * Attempt to log the user in via the provider user returned from Socialite.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  string  $provider
     * @return \Illuminate\Routing\Pipeline
     */
    public function handleProviderCallback(Request $request, string $provider, ResolvesSocialiteUsers $resolver)
    {
        if ($request->has('error')) {
            return Auth::check()
                ? redirect(config('fortify.home'))->dangerBanner($request->error_description)
[...]

PR #641, opened 2021-03-19
fix: HubSpot compatibility problem with joelbutcher/socialstream

nasatome:
The database table "connected_accounts" in joelbutcher/socialstream require a name and function "findConnectedAccountForProviderAndId" requires an ID user from provider.

fix: https://github.com/SocialiteProviders/Providers/issues/640

### DIFF
--- a/src/HubSpot/Provider.php
+++ b/src/HubSpot/Provider.php
@@ -58,7 +58,7 @@ class Provider extends AbstractProvider
         return (new User())->setRaw($user)->map([
             'nickname' => null, 'name' => '',
             'email'    => $user['user'], 'avatar' => null,
-            'id'       => $user['user_id']
+            'id'       => $user['user_id'],
         ]);
     }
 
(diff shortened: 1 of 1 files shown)


nasatome:
And I know that it depends on the database that the person uses but, I found out that NULL vs "" is insignificant in terms of disk-space and performance.

I do not consider that there is any break change, since the name normally in Laravel is only "printed" and in JavaScript the treatment that will be given will be practically the same if it changes the data from null to "" and the if(null || "") is the same result.... or not?

atymic:
> I do not consider that there is any break change, since the name normally in Laravel is only "printed" and in JavaScript the treatment that will be given will be practically the same if it changes the data from null to "" and the if(null || "") is the same result.... or not?

I think we should leave it as null, as not all providers return a name.

Apart from that, LGTM :)

src/HubSpot/Provider.php
```diff
@@ -56,8 +59,9 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'nickname' => null, 'name' => null,
+            'nickname' => null, 'name' => '',
```
atymic, line 62: Feel free to put these one per line so it's easier to see :)

nasatome:
@atymic How do you see the change?

atymic (approved):

PR #644, opened 2021-03-24
[Apple] Add support for a `user` as an associative array in Apple provider

### LINKED ISSUE #643: [Apple] Don't JSON-decode user from request if it's already an array
I'm currently using the Apple provider for social login from a mobile app, which communicates with a Laravel API at the back-end. The app authenticates with Apple first, and I verify with Apple (with Socialite) that the app is sending valid IDs and name, before persisting the state in a DB.

I'm facing an issue whereas at https://github.com/SocialiteProviders/Providers/blob/master/src/Apple/Provider.php#L205 the `user` is already a JSON-decoded associative array, which makes Socialite crash at that line with an error `Array to string conversion`

I was wondering if you'd be open for a PR that checks that the `user` coming through is an array before casting, happy to open it if you guys think it can help.

Apologies if this is not the right place to open, I'm not very good at OAuth in general 😓

thtg88:
I'm currently using the Apple provider for social login from a mobile app, which communicates with a Laravel API at the back-end.
The app authenticates with Apple first, and I verify with Apple (with Socialite) that the app is sending valid IDs and name, before persisting the state in a DB.

I'm facing an issue whereas the Apple provider at https://github.com/SocialiteProviders/Providers/blob/master/src/Apple/Provider.php#L205 tries to cast to string (to be later JSON-decoded) an associative array, which makes Socialite crash at that line with an error `Array to string conversion`

This PR aims at catching that case, together with isolating the logic for getting the user from the request.

Let me know if there is anything I need/can do in terms of CI/tests :)

Closes #643

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -202,19 +202,15 @@ class Provider extends AbstractProvider
      */
     protected function mapUserToObject(array $user)
     {
-        $value = trim((string) $this->request->input('user'));
+        $userRequest = $this->getUserRequest();
 
-        if ($value !== '') {
-            $userRequest = json_decode($value, true);
-
-            if (isset($userRequest['name'])) {
-                $user['name'] = $userRequest['name'];
-                $fullName = trim(
-                    ($user['name']['firstName'] ?? '')
-                    .' '
-                    .($user['name']['lastName'] ?? '')
-                );
-            }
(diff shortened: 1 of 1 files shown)


atymic (approved):

thtg88:
Thank you! :)

PR #645, opened 2021-03-26
Fixed invalid_auth error

brianwozeniak:


src/Slack/Provider.php
```diff
@@ -102,7 +102,12 @@ protected function getUserByToken($token)
     {
         try {
             $response = $this->getHttpClient()->get(
-                'https://slack.com/api/users.identity?token='.$token
+                'https://slack.com/api/users.identity',
+                [
+                    'headers' => [
+                        'Authorization' => 'Bearer '.$token,
```
atymic, line 108: Link to the docs please for this change? (but good to not pass it in the query :))

brianwozeniak:
Didn't review any docs regarding this. It simply does not work right now, and I copied how it was one on the majority of other extensions here to solve the issue. Confirmed that this fix works.

atymic (approved):

PR #646, opened 2021-03-28
Do not export tools folder.

lucasmichot:
It is useless to export the `tools` folder.


atymic (approved):
I have no idea what this does but you're prob right :)

PR #647, opened 2021-03-28
Do not allow class property typing.

lucasmichot:
Do not allow class property typing!
This feature was only introduced with PHP7.4 and this package supports PHP7.2


atymic:
It's actually 7.4
https://github.com/SocialiteProviders/Providers/blob/master/src/Adobe/composer.json

atymic:
The question of if we should make them all 7.2 is another thing. Since 7.2 is EOL already, I don't see a problem with having new providers as 7.4

lucasmichot:
> The question of if we should make them all 7.2 is another thing. Since 7.2 is EOL already, I don't see a problem with having new providers as 7.4

I agree with this, but for consistency and maintainability sake - I would ensure that all provides have the same php version constraints

PR #651, opened 2021-03-28
Normalize all composer.json files.

lucasmichot:
[Normalize](https://github.com/ergebnis/composer-normalize) all composer.json files.


atymic (commented):
Is there a lint command we can add to CI for this?

lucasmichot:
Not at the moment, but I guess that I could create tool/GA for that ?
This could come with another PR

lucasmichot:
The conflicts have just been fixed @atymic

PR #653, opened 2021-03-28
Decode the response body content.

lucasmichot:
We want to decode the response body content, not the response body.


atymic (approved):
Good catch. This will generally work as long as you don't try to decode it twice (it will work the first time, but the second time the pointer will be at the end and it will fail).

PR #658, opened 2021-03-28
Fixed Flickr doesn't recognise the permission set

brianwozeniak:
Currently when using Flickr it will result in the error: "Oops! Flickr doesn't recognise the permission set." - The authorize URL must have perms specified on it:

https://www.flickr.com/groups/51035612836@N01/discuss/72157692896948105/

Docs are here:

https://www.flickr.com/services/api/auth.oauth.html

I have verified that once this is added everything works correctly again.

### DIFF
--- a/src/Flickr/Server.php
+++ b/src/Flickr/Server.php
@@ -33,7 +33,7 @@ class Server extends BaseServer
         $authorizeUrl = 'https://www.flickr.com/services/oauth/authorize';
 
         if ($perms = $this->getConfig('perms')) {
-            return $authorizeUrl."?perms={$perms}";
+            return "{$authorizeUrl}?perms={$perms}";
         }
 
         return $authorizeUrl;
--- a/src/Flickr/Server.php
+++ b/src/Flickr/Server.php
@@ -22,7 +22,9 @@ class Server extends BaseServer
      */
     public function urlAuthorization()
     {
-        return 'https://www.flickr.com/services/oauth/authorize';
+        $perms = env('FLICKR_AUTHORIZATION_PERMS');
+
+        return 'https://www.flickr.com/services/oauth/authorize'.($perms ? "?perms={$perms}" : '');
     }
 
     /**
--- a/src/Flickr/Server.php
+++ b/src/Flickr/Server.php
@@ -12,6 +12,14 @@ class Server extends BaseServer
     /**
      * {@inheritdoc}
      */
+    public static function additionalConfigKeys()
+    {
+        return ['perms'];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function urlTemporaryCredentials()
     {
         return 'https://www.flickr.com/services/oauth/request_token';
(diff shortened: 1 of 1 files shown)


atymic:
"Additionally, you can pass the optional perms= parameter, asking for read, write, or delete privileges. This parameter will override the setting defined in your application's authentication flow."

Sounds like you should specify in your config on their end?

brianwozeniak:
Doesn’t work there, tried that. Feel free to test for yourself if you want.

Brian

atymic:
Just want to make sure we don't break anyone who has write perms set, and this overrides that.

How about we add it a config option, not set by default?

brianwozeniak:
Unfortunately Flickr doesn’t allow you to set anything, even though their docs say you can. However, I think a config option sounds like a great idea, especially if Flickr fixes their system then both would work, and you have the option to choose which permission set you want via both. It’s weird they don’t let you specify that as scopes like most.

atymic:
Sure, could you add the config var to the PR and set it to null by default? Thanks!

brianwozeniak:
Okay this should be good to go now using an `env` variable to be consistent with how everything is done in this repository. Let me know if you see any issues.

src/Flickr/Server.php
```diff
@@ -22,7 +22,9 @@ public function urlTemporaryCredentials()
      */
     public function urlAuthorization()
     {
-        return 'https://www.flickr.com/services/oauth/authorize';
+        $perms = env('FLICKR_AUTHORIZATION_PERMS');
```
atymic, line 25: Please use `additionalConfigKeys` and `getConfig` and update the readme

brianwozeniak: Unfortunately, I do not see any documentation on utilizing this:

https://socialiteproviders.com/contribute/

Perhaps I missed where this is?

Additionally, I looked through some other providers using these methods and am failing to see documentation there that might provide some clues on exactly what you are wanting. I simply looked through numerous of the providers here and saw that `env` was used directly, so followed by example.

atymic:
Sorry, could be documented better. We shouldn't use `env()` as it breaks config caching. I'll create an issue for it.

See auth0 as an example. To add keys:
https://github.com/SocialiteProviders/Providers/blob/e4a979bc081876bbcc37f74fa32ae3657bb31c00/src/Auth0/Provider.php#L35-L38

To use:
https://github.com/SocialiteProviders/Providers/blob/e4a979bc081876bbcc37f74fa32ae3657bb31c00/src/Auth0/Provider.php#L29

lucasmichot:
I agree that some documentation is lacking.
@brianwozeniak feel free to merge this PR: https://github.com/brianwozeniak/Providers/pull/1

lucasmichot:
The PR has been updated @atymic

brianwozeniak:
Thanks @lucasmichot, merged yours in. @atymic I believe this is ready to go now, please check.

atymic:
Actually, could you also add to the readme please? New PR is fine, then i'll tag new vers.

PR #661, opened 2021-04-03
Stop using env() helper.

### LINKED ISSUE #660: Remove usage of `env()` from providers
See https://laravel.com/docs/8.x/configuration#configuration-caching

These providers will break if `config:cache` is ran.

We should update them to use the `config()` helper, or the `getConfig` and `Provider::additionalConfigKeys()` method.

lucasmichot:
Stop using env() helper.

The provider can break if `config:cache` is run.

---

Fix #660


atymic (approved):

atymic:
Reason I didn't merge this is this will break existing installs using these parms that don't have them set as config options. I guess we need a major version bump for all of these providers?

PR #667, opened 2021-04-12
Lint the PHP code

lucasmichot:
Lint the PHP code for all the versions of PHP

This is very required as we have no tests running for this repository.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,13 +3,14 @@ name: Test
 on: [ pull_request, push ]
 
 jobs:
-  run:
+  test:
     runs-on: ubuntu-20.04
     strategy:
       max-parallel: 15
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


atymic:
We have styleci running, what's the reason for this sorry?

lucasmichot:
> We have styleci running, what's the reason for this sorry?

StyleCI is only checking for the code style - and reports errors. The service runs on PHP8.
It uses php-cs-fixer (+ custom) rules to ensure that all versions of recent PHP versions can be covered.

But at no time is code linting executed.
if we were to run tests for PHP versions, those errors in the code would break the execution.
Linting the code is a fast way to detect them.

atymic:
Can you resolve the issues please :)

atymic (changes requested):
Fix conflicts

lucasmichot:
> Fix conflicts

Done @atymic

atymic (approved):

PR #668, opened 2021-04-13
Simplify google provider.

lucasmichot:
This PR simplifies the Google provider.


atymic:
@m1guelpf do you remember why we re-implemented this provider?

Issue #670, opened 2021-04-27
Documentation issue for the SalesForce driver


nwoodhouse:
The documentation found here: https://socialiteproviders.com/SalesForce/#installation-basic-usage appears to have a small typo. It directs that the name of the driver configured in services.php should be:

'SalesForce'

However - this appears to result in a driver not supported error. Instead, the name of the driver ought to be:

'salesforce'

Should be a simple fix to update.

atymic:
Feel free to PR :)

nwoodhouse:
> Feel free to PR :)

Happy to - but I can't find the documentation repo any place :-/

atymic:
> > Feel free to PR :)
> 
> Happy to - but I can't find the documentation repo any place :-/

It's in this repo :)
https://github.com/SocialiteProviders/Providers/blob/master/src/SalesForce/README.md

nwoodhouse:
Done :-). My first pull request for an open source project - wahoo :-)

atymic:
Closed in #671

PR #671, opened 2021-04-28
Correcting Salesforce driver name in documentation

nwoodhouse:
As per: https://github.com/SocialiteProviders/Providers/issues/670#issuecomment-828375326, the name of the driver should be 'salesforce'. Using 'SalesForce' results in a driver not supported error message.


atymic (approved):
Thanks :)

nwoodhouse:
Awesome! Feels great to have a correction merged!

PR #672, opened 2021-05-03
Removed extra slash from Okta token URL when auth_server_id config value is null

kaypro4:
The following error is thrown when exchanging a code created at the standard Okta Authorization Server because the URLs don't match.

`https://dev-ID.okta.com/oauth2/v1` VS `https://dev-ID.okta.com/oauth2//v1`

111229386-6f693600-85ab-11eb-9c26-36daaba29484

The other (authorize and userinfo) endpoints already handle this conditional inclusion of the slash properly.

Thanks

### DIFF
--- a/src/Okta/Provider.php
+++ b/src/Okta/Provider.php
@@ -81,7 +81,7 @@ class Provider extends AbstractProvider
      */
     protected function getTokenUrl()
     {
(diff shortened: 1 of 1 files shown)


lucasmichot:
Fix comes in https://github.com/SocialiteProviders/Providers/pull/674

PR #674, opened 2021-05-04
Fix OKTA token URL.

lucasmichot:
This PR attempts to fix https://github.com/SocialiteProviders/Providers/pull/672
This typically happens when no `auth_server_id` has been set

### DIFF
--- a/src/Okta/Provider.php
+++ b/src/Okta/Provider.php
@@ -61,6 +61,16 @@ class Provider extends AbstractProvider
     }
 
     /**
+     * Get the Okta sever URL.
+     *
+     * @return string
+     */
+    protected function getOktaServerUrl(): string
+    {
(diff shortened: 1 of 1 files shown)


src/Okta/Provider.php
```diff
@@ -51,13 +51,19 @@ protected function getOktaUrl()
      */
     protected function getAuthServerId()
     {
-        $auth_server_id = $this->getConfig('auth_server_id', null);
+        $authServerId = (string) $this->getConfig('auth_server_id');
 
-        if ($auth_server_id) {
-            return $auth_server_id.'/';
-        }
+        return $authServerId === '' ? $authServerId : $authServerId.'/';
```
atymic, line 56: tbh i'd prefer an empty() here, but i don't see how the current logic works 🤷

atymic (approved):

Issue #681, opened 2021-05-31
Stripe - Undefined array key "email" when email is not returned in /v1/account call


franjohn21:
I'm using the Stripe provider and at least in local development in test mode I am able to see Stripe does not always return an email for the user. 

This is consistent with Stripe's documentation that the `email` field can be `null` for the `Account` object ([code](https://github.com/stripe/stripe-php/blob/master/lib/Account.php#L27)).

Because we hardcode `user["email"]` - the socialite provider throws an error in this case `Undefined array key "email"`. 

I can open up a pull request to fix this as it should be easy to make this optional. Just to confirm though - is this library production ready? I would figure something like this would have been seen already.

atymic:
Tagged in https://github.com/SocialiteProviders/Stripe/releases/tag/4.1.1

PR #682, opened 2021-05-31
Fix Stripe "email" error

franjohn21:
Fixes https://github.com/SocialiteProviders/Providers/issues/681 - missing "email" in account API response.

### DIFF
--- a/src/Stripe/Provider.php
+++ b/src/Stripe/Provider.php
@@ -64,7 +64,6 @@ class Provider extends AbstractProvider
         } elseif (isset($user['display_name'])) { // original location
             $nickname = $user['display_name'];
         }
-        
         $email = null;
         if (isset($user['email'])) {
             $email = $user['email'];
--- a/src/Stripe/Provider.php
+++ b/src/Stripe/Provider.php
@@ -70,8 +70,11 @@ class Provider extends AbstractProvider
         }
 
         return (new User())->setRaw($user)->map([
-            'id'   => $user['id'], 'nickname' => $nickname,
-            'name' => null, 'email' => $user['email'] ?? null, 'avatar' => null,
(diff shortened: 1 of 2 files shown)


src/Stripe/Provider.php
```diff
@@ -64,10 +64,15 @@ protected function mapUserToObject(array $user)
         } elseif (isset($user['display_name'])) { // original location
             $nickname = $user['display_name'];
         }
+
```
franjohn21, line 67:
```suggestion
```

src/Stripe/Provider.php
```diff
@@ -64,10 +64,14 @@ protected function mapUserToObject(array $user)
...
             $nickname = $user['display_name'];
         }
+        $email = null;
+        if (isset($user['email'])) {
+            $email = $user['email'];
+        }
 
         return (new User())->setRaw($user)->map([
             'id'   => $user['id'], 'nickname' => $nickname,
-            'name' => null, 'email' => $user['email'], 'avatar' => null,
```
atymic, line 70: Could you split these props onto one per line.

src/Stripe/Provider.php
```diff
@@ -64,10 +64,14 @@ protected function mapUserToObject(array $user)
...
+            'name' => null, 'email' => $email, 'avatar' => null,
```
atymic, line 74:
```suggestion
            'name' => null, 'email' => $user['email'] ?? null, 'avatar' => null,
```

atymic (changes requested):

atymic:
Looks good just use null coalesce instead of the conditional.

atymic (commented):
Could you also update the docs to note the email field is optional please?

atymic (approved):

PR #689, opened 2021-06-16
Update Constant Contact to use V3 endpoints and scopes

jstowey:
Constant Contact changed all their endpoints for version 3 of the API so this fixes that.


atymic:
Thanks!

PR #694, opened 2021-07-05
Added TikTok provider

tombanksme:
Added a [TikTok Login Kit](https://developers.tiktok.com/doc/login-kit-web) provider.

### DIFF
--- a/src/TikTok/Provider.php
+++ b/src/TikTok/Provider.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace SocialiteProviders\Facebook;
+namespace SocialiteProviders\TikTok;
 
 use Illuminate\Support\Arr;
 
@@ -20,7 +20,7 @@ class Provider extends AbstractProvider
      * {@inheritdoc}
      */
     protected $scopes = [
-        'user.info.basic'
+        "user.info.basic"
     ];
 
     /**
@@ -59,6 +59,14 @@ class Provider extends AbstractProvider
         return $this->user->setToken($token)
             ->setExpiresIn(Arr::get($response, "data.expires_in"))
             ->setRefreshToken(Arr::get($response, "data.refresh_token"));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTokenUrl()
+    {
+        return "https://open-api.tiktok.com/oauth/access_token/";
     }
 
     /**
--- a/src/TikTok/Provider.php
+++ b/src/TikTok/Provider.php
@@ -4,10 +4,9 @@ namespace SocialiteProviders\TikTok;
 
 use Illuminate\Support\Arr;
 
-use SocialiteProviders\Manager\OAuth2\User;
-use SocialiteProviders\Manager\OAuth2\AbstractProvider;
-
 use Laravel\Socialite\Two\InvalidStateException;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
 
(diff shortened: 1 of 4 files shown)


src/TikTok/Provider.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace SocialiteProviders\Facebook;
```
Grasume: Update this

src/TikTok/Provider.php
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Unique Provider Identifier.
+     */
+    public const IDENTIFIER = 'TIKTOK';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = [
+        "user.info.basic"
```
atymic, line 23: single quotes please

src/TikTok/Provider.php
```diff
@@ -0,0 +1,122 @@
...
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state)
+    {
+        return "https://open-api.tiktok.com/platform/oauth/connect?" . http_build_query([
```
atymic, line 31: also here.

src/TikTok/Provider.php
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Get TikTok user by token.
+     * 
+     * @param array $data
+     * @return mixed
+     */
+    protected function getUserByToken($data)
+    {
+        // Note: The TikTok api does not currently have a way to get the user data
+        // with only the access token.
```
atymic, line 96: Not sure what you mean here?

src/TikTok/Provider.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $response = $this->getHttpClient()->get(
+            "https://open-api.tiktok.com/oauth/userinfo?" . http_build_query([
+                "open_id" => $data[1],
+                "access_token" => $data[0]
```
atymic, line 101: Is is possible not to pass this through the query string?

atymic (changes requested):

src/TikTok/Provider.php
tombanksme: The TikTok API does not have an endpoint where it returns the authorized user based on the access token only. You also have to give it the open-id to get the user information.

Thats why I've had to change $data to be an array where $data[0] is the access token and $data[1] is the user's open_id.

src/TikTok/Provider.php
tombanksme: I'm not too sure what you mean? If it's about the use of http_build_query I just think it looks nicer, I can remove it and just use string composition instead if you would prefer.

src/TikTok/Provider.php
```diff
@@ -0,0 +1,124 @@
...
+        if ($this->hasInvalidState()) {
+            throw new InvalidStateException();
+        }
+
+        $response = $this->getAccessTokenResponse($this->getCode());
+
+        $token = Arr::get($response, 'data.access_token');
+
+        $this->user = $this->mapUserToObject(
+            $this->getUserByToken([$token, Arr::get($response, 'data.open_id')])
```
atymic, line 56: Can we key these array params? Just for the sake of readability as `'open_id'      => $data[1],` would be easier to interpret as `'open_id'      => $data['open-id'],`

 :) Thanks

atymic (commented):
Just one minor change then good to go, sorry!

atymic (approved):

PR #695, opened 2021-07-07
Normalize composer files

lucasmichot:
Normalize come `composer.json` files that have been recently added/updated.


atymic:
Thanks!

PR #691, opened 2021-06-21
Create Eveonline 0uath2 provider

Grasume:
This Pr is for the addition of the Eveonline provider.

PR #696, opened 2021-07-07
Eve online

lucasmichot:
This PR fixes lots of inconsistencies and issues with EveOnline provider that was introduced in https://github.com/SocialiteProviders/Providers/pull/691

---

See the commits for more details.


atymic (approved):
Thanks @lucasmichot!

PR #697, opened 2021-07-07
Fixed undefined field for Rekono

nunar:
Some users do not have information for field `name`.


nunar (commented):
Typo

src/Rekono/Provider.php
```diff
@@ -67,6 +67,10 @@ protected function getUserByToken($token)
      */
     protected function mapUserToObject($user)
     {
+        if (!isset($user['name'])) {
+            $user['name'] = null;
+        }
+
         return (new User())->setRaw($user)->map([
             'id'    => $user['sub'],
             'name'  => $user['name'],
```
lucasmichot, line 76: Why not just `'name'  => $user['name'] ?? null,` ?

nunar: 👍 Thank you!

atymic (approved):

PR #699, opened 2021-07-07
Consistant configuration access.

lucasmichot:
Enforce consistancy for configuration access.
We shall always use getConfig for the keys which are related to the service.

### DIFF
--- a/src/Ovh/Provider.php
+++ b/src/Ovh/Provider.php
@@ -22,7 +22,10 @@ class Provider extends AbstractProvider
     public static function additionalConfigKeys()
     {
         return [
+            'client_id',
+            'client_secret',
             'endpoint',
+            'redirect',
         ];
     }
 
--- a/src/Ovh/Provider.php
+++ b/src/Ovh/Provider.php
@@ -17,16 +17,11 @@ class Provider extends AbstractProvider
     public const IDENTIFIER = 'OVH';
 
     /**
-     * @return array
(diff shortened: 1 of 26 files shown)


src/Ovh/Provider.php
```diff
@@ -22,7 +22,10 @@ class Provider extends AbstractProvider
     public static function additionalConfigKeys()
     {
         return [
+            'client_id',
+            'client_secret',
```
atymic, line 26: these are default right?

src/Trakt/Provider.php
```diff
@@ -75,6 +75,6 @@ protected function getTokenFields($code)
      */
     public static function additionalConfigKeys()
     {
-        return ['api_version'];
+        return ['api_version', 'client_id'];
```
atymic, line 78: Think this is also covered by the default config items

atymic (commented):
Looks good, few minor bits :)

src/Ovh/Provider.php
lucasmichot: Good catch - see 565ae93ad50d6f97970000878c940f3e0eb4b707

src/Trakt/Provider.php
lucasmichot: See 565ae93ad50d6f97970000878c940f3e0eb4b707

atymic (approved):

PR #704, opened 2021-07-14
Explicitly decode the body

lucasmichot:
Explicitly decode the body content

### DIFF
--- a/src/Atlassian/Provider.php
+++ b/src/Atlassian/Provider.php
@@ -52,7 +52,7 @@ class Provider extends AbstractProvider
             ],
         ]);
 
-        return json_decode($response->getBody()->getContents(), true);
+        return json_decode((string) $response->getBody(), true);
     }
 
     /**
--- a/src/Atlassian/Provider.php
+++ b/src/Atlassian/Provider.php
@@ -52,7 +52,7 @@ class Provider extends AbstractProvider
             ],
         ]);
 
-        return json_decode($response->getBody(), true);
+        return json_decode($response->getBody()->getContents(), true);
     }
 
     /**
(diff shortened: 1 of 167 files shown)


src/Atlassian/Provider.php
```diff
@@ -52,7 +52,7 @@ protected function getUserByToken($token)
             ],
         ]);
 
-        return json_decode($response->getBody(), true);
+        return json_decode($response->getBody()->getContents(), true);
```
atymic, line 55: Interestingly, I think the correct way to do this is `(string) $response->getBody()`

Reason being that getcontents reads from the current pointer position, but if anything (middleware, double call, etc) has called get contents before this point it will return the wrong thing, whereas casting to string reads the whole thing.

Passing it to json decode is an implicit cast to string.

```
/**
     * Reads all data from the stream into a string, from the beginning to end.
     *
     * This method MUST attempt to seek to the beginning of the stream before
     * reading data and read the stream until the end is reached.
     */
    public function __toString();

/**
     * Returns the remaining contents in a string
     *
     * @return string
     * @throws \RuntimeException if unable to read or an error occurs while
     *     reading.
     */
    public function getContents();
```

lucasmichot: I have actually never thought about those middleware.

lucasmichot: I this case I think that actually all decode should be consistant

lucasmichot: See ff6269374cac1da3de5a8fa5f336c0b96090687f @atymic

atymic (approved):

Issue #705, opened 2021-07-15
Saml2 Provider - custom SP entity id


peppeocchi:
Hi, I was wondering if there is any way to manually set the SP entity id. By looking at the code [here](https://github.com/SocialiteProviders/Saml2/blob/master/Provider.php#L221) it uses `URL::to('auth/saml2')` as entity id.
It would be great if that could be customised.

I'm now integrating with a IDP and the developer I'm talking to sent me this message
>  I’ve seen issues with apps that use the same entity ID for all implementations, and that causes issues when dealing with multiple environments and the ability to support multiple IDPs.

Did I miss any configuration or function to set a custom SP entity id?

Thanks

atymic:
Feel free to PR :D

27pchrisl:
This makes sense according to the spec. The implementation as it is sort of assumes each instance of your app (eg each customer) is on their own subdomain, causing different entity ids for different customers. If you don’t partition your app in the url like this you definitely need to be able to generate entity ids like (eg) http://app.com/auth/saml2/<customer id> so you know which customer to route an incoming assertion to.

atymic:
Close in #707

PR #707, opened 2021-07-16
Saml2 - Custom Service Entity Id

peppeocchi:
Ref https://github.com/SocialiteProviders/Providers/issues/705

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -92,6 +92,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
             'entityid',
             'certificate',
             'sp_acs',
+            'sp_entityid',
         ];
     }
 
@@ -218,7 +219,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
     {
         $entityDescriptor = new EntityDescriptor();
         $entityDescriptor
-            ->setEntityID(URL::to('auth/saml2'))
+            ->setEntityID($this->getConfig('sp_entityid', URL::to('auth/saml2')))
             ->addItem(
                 (new SpSsoDescriptor())
                     ->setWantAssertionsSigned(true)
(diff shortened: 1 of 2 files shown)


src/Saml2/Provider.php
```diff
@@ -218,7 +219,7 @@ public function getServiceProviderEntityDescriptor(): EntityDescriptor
     {
         $entityDescriptor = new EntityDescriptor();
         $entityDescriptor
-            ->setEntityID(URL::to('auth/saml2'))
+            ->setEntityID($this->getConfig('sp_entityid', URL::to('auth/saml2')))
```
atymic, line 222: Hmm, what does `URL::to('auth/saml2')` produce here? 
Should the entity id be inside the url to?

peppeocchi: Good point, I should probably change the [documentation](https://github.com/SocialiteProviders/Providers/pull/707/files#diff-a3aac639db6923f75aefb2cf21a84fc924dba38f3bb97659cf1748b2ffc41518R158) to look like a URI
```
'saml2' => [
  'metadata' => 'https://idp.co/metadata/xml',
  'sp_entityid' => 'https://my.domain.com/my/custom/entityid',
],
```

from the SAML2 Specification
> ... The syntax of such an identifier is a URI of not more than 1024 characters in length. It is RECOMMENDED that a system entity use a URL containing its own domain name to identify itself.


The option might be to validate that the custom sp_entityid is a URI and if so use it, otherwise default to what's already there (`URL::to('auth/saml2')`).

For consistency, it might make sense to do the same as per the sp_acs, which allows you to set your own path
```
->setEntityID(URL::to($this->getConfig('sp_entityid', 'auth/saml2')))
```
 which in theory is enough in order to allow unique entity ids for different IDPs. Although for the acs to have its own domain makes sense, it is not "required" for the entity id so one might argue that this change won't really allow a custom entityid.

atymic: Ok, happy with you just updating the docs instead of having laravel autogen :)

atymic (approved):

PR #729, opened 2021-09-01
[Apple] Update README.md to note the client secret JWT expiry

jryd:
The post from Okta explains how to configure Sign In with Apple and get it working.

The issue is that the statically generated JWT token that they use has a maximum lifetime of 6 months (a rule from Apple). I recently wrote about an approach off the back of the Okta article on generating the client secret for each request to ensure you never face an issue with the client secret expiring.


PirunSeng:
Excuse me @jryd  , could you share your environment please? I can't get the issue of "Driver apple not supported" fixed.
Here are my environments:
```
"require": {
        "php": "^7.4",
        "laravel/framework": "^8.0",
        "laravel/socialite": "^5.2",
        "socialiteproviders/apple": "^5.0"
    },
```
Thanks in advanced for your support :)

PR #733, opened 2021-09-14
Add Azure AD B2C Provider

fujie:
Add Azure AD B2C provider because OpenID Connect implementation on Azure AD B2C is bit different from Azure AD's one.

- Azure AD B2C does not issue access_token by default to obtain user attributes via Microsoft Graph API
- To obtain attributes from Azure AD B2C, it is required to get claims from id_token

### DIFF
--- a/src/AzureADB2C/Provider.php
+++ b/src/AzureADB2C/Provider.php
@@ -1,0 +1,153 @@
+<?php
+
+namespace SocialiteProviders\AzureADB2C;
+
+use Firebase\JWT\JWK;
+use GuzzleHttp\Client;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
+use Illuminate\Support\Facades\Cache;
+use Lcobucci\Clock\SystemClock;
+use Lcobucci\JWT\Configuration;
+use Lcobucci\JWT\Signer\Key\InMemory;
+use Lcobucci\JWT\Signer\Rsa\Sha256;
+use Lcobucci\JWT\Validation\Constraint\IssuedBy;
+use Lcobucci\JWT\Validation\Constraint\LooseValidAt;
+use Lcobucci\JWT\Validation\Constraint\SignedWith;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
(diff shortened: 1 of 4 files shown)


src/AzureADB2C/Provider.php
```diff
@@ -0,0 +1,171 @@
...
+    protected $scopes = [
+        'openid',
+    ];
+
+    /**
+     * Get OpenID Configuration and store on cache.
+     */
+    private function getOpenIdConfiguration()
+    {
+        return Cache::remember('socialite_'.self::IDENTIFIER.'_openidconfiguration', intval($this->config['cache_time'] ?: 3600), function () {
```
atymic, line 31: could we use sprintf for the cache key please?

atymic: does this work if cache time isn't set?

atymic (commented):
Looks good, just a couple of little bits :)

src/AzureADB2C/Provider.php
fujie: Thank you for your comment. Just I fixed to use sprintf here.
"?:" works fine for me if cache time isn't set, but to make this clear, I changed code to get default value from services.php.(described in README.md)

fujie:
@atymic just updated for your comments.

atymic:
@lucasmichot could you check this please? :)

fujie:
@lucasmichot thanks for your comments. just I modified the code.

fujie:
@lucasmichot Thank you for your review and comments! Just I fixed and could you please review again?

src/AzureADB2C/README.md
```diff
@@ -0,0 +1,69 @@
...
+```php
+return redirect(Socialite::driver('azureadb2c')->logout('http://localhost'));
+```
+
+### Returned User fields
+
+- ``sub``
+- ``name``
+
+Note) If you want to add claim mappings, change mapUserToObject function. The claims mappings must be match with claims in id_token which Azure AD B2C returns.
```
lsmith77, line 62: what is the recommended pattern to change the claim mapping?

PR #736, opened 2021-09-18
Add ability to request custom fields from Microsoft

mashedkeyboard:
This PR adds the ability to add application-specific requests for custom fields from the Microsoft provider. This allows for the use of fields like on-premise Active Directory extension attributes, which is my use case for it, but also any other field on the Microsoft Graph that an application might need to access.

### DIFF
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -15,6 +15,7 @@ class Provider extends AbstractProvider
 
     /**
      * Default field list to request from Microsoft.
+     * @see https://docs.microsoft.com/en-us/graph/permissions-reference#user-permissions
      */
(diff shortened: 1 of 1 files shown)


atymic (commented):
Does this require any additional scopes beyond the default?

mashedkeyboard:
No, it doesn't. The fields in DEFAULT_FIELDS are the same that would otherwise be used with the default scopes - so unless someone specifies fields, the behaviour hasn't changed.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

atymic:
Could you send over a link to the docs just so we can check please?

atymic:
ping @mashedkeyboard

mashedkeyboard:
Permissioning documentation from Microsoft is here, although it's not hugely helpful: https://docs.microsoft.com/en-us/graph/permissions-reference#user-permissions

PR #743, opened 2021-09-29
Update SoundCloud to newest security patch

jstowey:
SoundCloud updated their OAuth flow to use a token in the header, instead of passing it via the query string.

More info here: https://developers.soundcloud.com/blog/security-updates-api


atymic (approved):
Thanks!

PR #747, opened 2021-10-06
Always prefer RequestOptions::QUERY and http_build_query when possible.

lucasmichot:
Always prefer Req`uestOptions::QUERY` and `http_build_query` when possible.


atymic:
Can you rebase?

lucasmichot:
> Can you rebase?

Done @atymic 👌

PR #748, opened 2021-10-06
Directly use the request session.

lucasmichot:
Directly use the request session.


src/Apple/Provider.php
```diff
@@ -178,8 +178,10 @@ public function user()
         if ($this->usesState()) {
             $state = explode('.', $appleUserToken['nonce'])[1];
             if ($state === $this->request->input('state')) {
-                $this->request->session()->put('state', $state);
-                $this->request->session()->put('state_verify', $state);
+                $this->request->session()->put([
```
lucasmichot, line 181: No need to `put` twice

PR #760, opened 2021-10-28
added support for Linode

tomschlick:


src/Linode/Provider.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function mapUserToObject(array $user)
+    {
+        return (new User())->setRaw($user)->map([
+            'id'     => $user['uid'],
+            'name'   => $user['username'],
+            'email'  => $user['email'],
+            'avatar' => 'https://www.gravatar.com/avatar/'.md5($user['email']),
```
atymic, line 61: do we need to return this? I think it's prob best to just return null and someone can make a gravatar url if they want?

atymic (commented):
Just one minor thing

src/Linode/Provider.php
tomschlick: I’m fine with omitting it. It was copied from the DigitalOcean configuration.

atymic: Thanks :)

atymic (changes requested):
Could you please add it to the monorepobuilder.yml split script.

atymic (approved):

PR #767, opened 2021-11-08
Make SAML2 binding configurable & fix regression introduced by #712

coclav:
When implementing SAML2 configuration to work with jumpcloud.com as Identify Provider, I realised their SSO binding method is a "POST" and not a "REDIRECT". 

Problem: Redirect method is hardcoded in the code. 

Solution : This change makes it configurable in the services.php file. It is backward compatible as it takes the current value as default.

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -93,6 +93,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
             'certificate',
             'sp_acs',
             'sp_entityid',
+            'idp_binding_method'
         ];
     }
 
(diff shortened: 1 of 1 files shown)


coclav:
I'm also fixing a regression introduced by #712 which messed up the METADATA library, by using _HumbugBox58fd4d9e2a25\RdKafka\Metadata; (whatever this is ??) instead of the lightsaml


See bad changes here : 

https://github.com/SocialiteProviders/Providers/commit/c4f5d82e173e34b682ff6a4813db30fdf12fc406#diff-b28848b818f26254c5f0d5090662fff2f8406fe28bda138614247d01bab91a18

atymic:
Thanks
cc @lucasmichot

PR #772, opened 2021-11-17
Update ProductHunt API to v2

thinkverse:
This PR updates the ProductHunt driver to use the [v2](https://api.producthunt.com/v2/docs) API since v1 is now [deprecated](https://api.producthunt.com/v1/docs). This means the users' email is no longer returned, I've opened an [issue](https://github.com/producthunt/producthunt-api/issues/247) in ProductHunts' producthunt-api repo related to that since it was included in v1, so I've asked if it could be added to the v2 GraphQL [User Schema](http://api-v2-docs.producthunt.com.s3-website-us-east-1.amazonaws.com/object/user/). If it gets added I'll update the provider again. 

This PR also fixes the driver since the PSR-4 namespace was sadly wrong so Composer autoloaded the driver under the `SocialiteProviders\AngelList` namespace instead of `SocialiteProviders\ProductHunt` causing a ReflectionError when Laravel tried to resolve the driver. 🤷‍♂️

Thank you for making this package, an amazing addition to the Laravel community. 🤘🔥

### DIFF
--- a/src/ProductHunt/Provider.php
+++ b/src/ProductHunt/Provider.php
@@ -33,7 +33,7 @@ class Provider extends AbstractProvider
     protected function getAuthUrl($state)
     {
         return $this->buildAuthUrlFromBase(
-            'https://api.producthunt.com/v1/oauth/authorize',
+            'https://api.producthunt.com/v2/oauth/authorize',
             $state
         );
     }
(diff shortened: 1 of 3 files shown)


thinkverse:
Guess the style guide doesn't like heredoc. 😆

atymic (approved):
Thanks!

PR #781, opened 2021-12-22
Saml2: Extend SAML2 support

danielztolnai:
This pull request extends the Saml2 provider to support more of the SAML2 standard. Here is a list of all new functionality:
- Full backwards compatibility, **no** breaking changes
- **Singing and encryption**
  Many Identity Providers require the signing of requests and providing of a certificate for encryption. The provider can now take a certificate + private key as optional configuration parameters to support this. If they are provided, every outgoing message is signed, and the provider supports decrypting assertions coming from the provider. If the parameters are not provided there is no change in behaviour.
- **Support both `POST` and `REDIRECT` bindings simultaneously while using `POST` as the default**
  The current implementation prefers `REDIRECT` over `POST`. I added the option to support both and still be able to change the default. The change is backwards compatible as `REDIRECT` still takes precedence by default.
- **More information in the Service Provider metadata**
  Identity Providers can take more information from the metadata than provided by the current version. This PR adds the following extra information:
  - Every supported binding is present, not just the default one
  - The default binding method is marked as default
  - The required NameId format is present
  - Signing and encryption information when it is configured
  - Possibility to add technical contact details
  - Possibility to add organization information
- **Configurable attribute mapping with multiple accepted fields**
  - Support for both the `urn:oid:` and `http://schemas.xmlsoap.org/` namespaces
  - More fields are mapped automatically with complete backwards compatibility with the current version
  - Provider configuration parameter to override / extend the mappings
  - Popular `urn:oid:` attributes are available as a class of consts
- State checking is more conformant with the other providers and now supports the `stateless` feature
- Response status check is added to avoid unhandled errors trying to parse a message that only consists of an error status.
- **SAML2 Single Logout**
  SAML2 supports centralized logout. It comes with its drawbacks, but it is still good to have the option in the provider to support this feature. Warning about possible complications is in the readme.

Every new feature is documented in the readme.

The new features have been tested using Shibboleth, Azure, Microsoft ADFS and Okta.

Thank you for considering this extension.

### DIFF
--- a/src/Saml2/OasisAttributeNameUris.php
+++ b/src/Saml2/OasisAttributeNameUris.php
@@ -8,6 +8,7 @@ class OasisAttributeNameUris
     public const DISPLAY_NAME = 'urn:oid:2.16.840.1.113730.3.1.241';
     public const GIVEN_NAME = 'urn:oid:2.5.4.42';
     public const MAIL = 'urn:oid:0.9.2342.19200300.100.1.3';
+    public const ORGANIZATIONAL_UNIT_NAME = 'urn:oid:2.5.4.11';
     public const PHONE = 'urn:oid:2.5.4.20';
     public const PREFERRED_FIRST_NAME = 'urn:oid:1.2.840.113994.200.47';
     public const PREFERRED_MIDDLE_NAME = 'urn:oid:1.2.840.113994.200.48';
--- a/src/Saml2/OasisAttributeNameUris.php
+++ b/src/Saml2/OasisAttributeNameUris.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace SocialiteProviders\Saml2;
+
+class OasisAttributeNameUris
+{
+    public const COMMON_NAME = 'urn:oid:2.5.4.3';
+    public const DISPLAY_NAME = 'urn:oid:2.16.840.1.113730.3.1.241';
+    public const GIVEN_NAME = 'urn:oid:2.5.4.42';
+    public const MAIL = 'urn:oid:0.9.2342.19200300.100.1.3';
+    public const PHONE = 'urn:oid:2.5.4.20';
+    public const PREFERRED_FIRST_NAME = 'urn:oid:1.2.840.113994.200.47';
+    public const PREFERRED_MIDDLE_NAME = 'urn:oid:1.2.840.113994.200.48';
+    public const PREFERRED_SURNAME = 'urn:oid:1.2.840.113994.200.49';
+    public const REGISTERED_GIVEN_NAME = 'urn:oid:1.2.840.113994.200.32';
+    public const REGISTERED_SURNAME = 'urn:oid:1.2.840.113994.200.31';
+    public const SURNAME = 'urn:oid:2.5.4.4';
+    public const TITLE = 'urn:oid:2.5.4.12';
+    public const UID = 'urn:oid:0.9.2342.19200300.100.1.1';
+}
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -521,7 +521,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
             return;
         }
 
-        $assertion = $reader->decryptMultiAssertion([$credential], new DeserializationContext());
+        $assertion = $reader->decryptAssertion($credential->getPrivateKey(), new DeserializationContext());
         $this->messageContext->asResponse()->addAssertion($assertion);
     }
 
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -337,9 +337,9 @@ class Provider extends AbstractProvider implements SocialiteProvider
             ->getAllKeyDescriptorsByUse(KeyDescriptor::USE_SIGNING);
 
         /** @var SignatureXmlReader $signatureReader */
-        $signatureReader = SamlConstants::BINDING_SAML2_HTTP_REDIRECT === $this->getAssertionConsumerServiceBinding() ?
-            $this->messageContext->getMessage()->getSignature() :
-            $this->messageContext->asResponse()->getFirstAssertion()->getSignature();
+        $signatureReader = SamlConstants::BINDING_SAML2_HTTP_REDIRECT === $this->messageContext->getBindingType()
+            ? $this->messageContext->getMessage()->getSignature()
+            : $this->messageContext->asResponse()->getFirstAssertion()->getSignature();
 
         if (!$signatureReader) {
             return true;
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -236,18 +236,24 @@ class Provider extends AbstractProvider implements SocialiteProvider
 
     public function getServiceProviderEntityDescriptor(): EntityDescriptor
     {
+        $spSsoDescriptor = new SpSsoDescriptor();
+        $spSsoDescriptor->setWantAssertionsSigned(true);
+
+        foreach ([SamlConstants::BINDING_SAML2_HTTP_REDIRECT, SamlConstants::BINDING_SAML2_HTTP_POST] as $binding) {
+            $acsRoute = $this->getAssertionConsumerServiceRoute();
+            if ($this->hasRouteBindingType($acsRoute, $binding)) {
+                $spSsoDescriptor->addAssertionConsumerService(
+                    (new AssertionConsumerService())
+                        ->setBinding($binding)
+                        ->setLocation(URL::to($acsRoute))
+                );
+            }
+        }
+
         $entityDescriptor = new EntityDescriptor();
         $entityDescriptor->setID(Helper::generateID())
             ->setEntityID($this->getConfig('sp_entityid', URL::to('auth/saml2')))
-            ->addItem(
-                (new SpSsoDescriptor())
-                    ->setWantAssertionsSigned(true)
-                    ->addAssertionConsumerService(
-                        (new AssertionConsumerService())
-                            ->setBinding($this->getDefaultAssertionConsumerServiceBinding())
-                            ->setLocation(URL::to($this->getAssertionConsumerServiceRoute()))
-                    )
-            );
+            ->addItem($spSsoDescriptor);
 
         return $entityDescriptor;
     }
@@ -256,7 +262,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
     {
         return $this->getServiceProviderEntityDescriptor()
             ->getFirstSpSsoDescriptor()
-            ->getFirstAssertionConsumerService()
+            ->getFirstAssertionConsumerService($this->getDefaultAssertionConsumerServiceBinding())
             ->getLocation();
     }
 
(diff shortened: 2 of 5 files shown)


Magnie:
@atymic @lucasmichot
I happen to need some of these changes for one of my projects. Are there any changes that need to be made to this pull request before it can be merged in? Anything I can help with?

src/Saml2/Provider.php
```diff
@@ -110,28 +170,64 @@ public function redirect()
     {
         $this->request->session()->put('state', $state = $this->getState());
```
Magnie, line 171:
```suggestion
        $state = null;

        if ($this->usesState()) {
            $this->request->session()->put('state', $state = $this->getState());
        }
```

Magnie (commented):
As my co-worker and I were working to get stateless stuff working for SAML 2, we also encountered an error with the `redirect()` method failing because it also didn't have support for `stateless()` like you had added for `hasInvalidState()`.

atymic (approved):
Hope you had a good break! 

Looks good to me, @lucasmichot any issues?
Because of the large amount of changes i was thinking on a major version bump just to ensure we don't break any existing installs?

danielztolnai:
@Magnie is right, my update missed full stateless support in the `redirect()` function, so I added it. The PR should still be backwards compatible as the `$stateless` flag is `false` by default (and using `stateless()` is not working in the current production version)

src/Saml2/Provider.php
danielztolnai: Added a slightly different code so the `null` state is not appended at all.

src/Saml2/README.md
```diff
@@ -166,12 +211,43 @@ Socialite::driver('saml2')->getServiceProviderEntityId()
 Socialite::driver('saml2')->getServiceProviderAssertionConsumerUrl()
 ```
 
+You can also publish the service provider's organization and technical support contact in the metadata by configuring them in `config/services.php` as:
+```php
+'saml2' => [
```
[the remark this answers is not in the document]
danielztolnai: These are not mandatory fields in the standard and there is no functionality lost if they are not present in the metadata. Therefore, this provider does not require them to be configured and I left them empty by default so the fields will not even appear in the metadata.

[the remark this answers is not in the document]
27pchrisl: Got it, I misunderstood line 227 of the README about those fields being "required". You're saying it doesn't mean they're required to have a working configuration of the provider, it means they're required if you want to use the functionality those fields are involved in.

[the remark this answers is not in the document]
danielztolnai: Yes, exactly. Sorry about the confusion. I'll try to clarify it in the readme.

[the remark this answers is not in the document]
27pchrisl: Yup, makes sense thanks!

src/Saml2/Provider.php
```diff
@@ -274,69 +418,77 @@ public function user()
...
-        $assertion = $messageContext->asResponse()->getFirstAssertion();
+        $assertion = $this->messageContext->asResponse()->getFirstAssertion();
         $attributeStatement = $assertion->getFirstAttributeStatement();
 
         $this->user = new User();
         $this->user->setAssertion($assertion);
         $this->user->map(['id' => $assertion->getSubject()->getNameID()->getValue()]);
 
         if ($attributeStatement) {
-            foreach (['name' => ClaimTypes::NAME, 'email' => ClaimTypes::EMAIL_ADDRESS] as $key => $claim) {
```
Jellyfrog, line 299: Nice PR!

I was just preparing to submit a PR extending support for the attributes, Google is not sending `NAME` but sends `GIVEN_NAME` and `SURNAME` [by default].
So I made a simple patch that (tries) to build `name` from these attributes...
 
After reading up about `NAME` I realized it's rather misunderstood, Microsoft states:
>The unique name of the user

Found another person saying:
>The NameID is indeed commonly used for the accountname, but can also contain the emailaddress(when the emailaddress format is used) or other unique attributes. you can give the NameID claim specific formats such as "unspecified" , "emailaddress", "persistant" etc.

>I would not recommend to use the NameID for names(firstname, lastname etc.). It's ment as holder for the unique attribute what is used by the relying party to uniquely identify the user.

https://social.msdn.microsoft.com/Forums/vstudio/en-US/1d36671f-fcb7-48b5-b590-82c6e125af66/correct-usage-of-the-name-claim-httpschemasxmlsoaporgws200505identityclaimsname?forum=Geneva

Thoughts?

Jellyfrog: Here is the patch for reference;
```diff
diff --git a/Provider.php b/Provider.php
index afd40a7..d003b86 100644
--- a/Provider.php
+++ b/Provider.php
@@ -304,6 +304,16 @@ public function user()
                     $this->user->map([$key => $attribute->getFirstAttributeValue()]);
                 }
             }
+
+            if (!$this->user->name) {
+                if ($attribute = $attributeStatement->getFirstAttributeByName(ClaimTypes::GIVEN_NAME)) {
+                    $this->user->name = $attribute->getFirstAttributeValue();
+                }
+
+                if ($attribute = $attributeStatement->getFirstAttributeByName(ClaimTypes::SURNAME)) {
+                    $this->user->name = trim($this->user->name . " " . $attribute->getFirstAttributeValue());
+                }
+            }
 
             $this->user->setRaw($attributeStatement->getAllAttributes());
         }
```

danielztolnai: NameID (read from the assertion [here](https://github.com/SocialiteProviders/Providers/pull/781/files#diff-b28848b818f26254c5f0d5090662fff2f8406fe28bda138614247d01bab91a18R439)) and `ClaimTypes::NAME` (`http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name`) are not necessarily the same.

The first one always has to be present and this provider requires a persistent NameID from the IdP to make sure it can be used as a unique identifier.

On the other hand the claim `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name` is just one of the many claims that can be sent and is not necessarily present in the data at all.

Some IdPs can send the same data for the two, that is right, but they can also send different values.

I think we have to keep `ClaimTypes::NAME` as the first option processed by default as the original provider code uses it as well and we would break backwards compatibility if we changed it.

However, with the new configuration options, you can simply remove it in your `config/services.php` like this:
```php
'saml2' => [
    'attribute_map' => [
        'name' => [
            OasisAttributeNameUris::DISPLAY_NAME,
            ClaimTypes::COMMON_NAME,
            OasisAttributeNameUris::COMMON_NAME,
        ],
    ],
],
```

src/Saml2/OasisAttributeNameUris.php
```diff
@@ -0,0 +1,21 @@
...
+    public const COMMON_NAME = 'urn:oid:2.5.4.3';
+    public const DISPLAY_NAME = 'urn:oid:2.16.840.1.113730.3.1.241';
+    public const GIVEN_NAME = 'urn:oid:2.5.4.42';
+    public const MAIL = 'urn:oid:0.9.2342.19200300.100.1.3';
+    public const ORGANIZATIONAL_UNIT_NAME = 'urn:oid:2.5.4.11';
+    public const PHONE = 'urn:oid:2.5.4.20';
+    public const PREFERRED_FIRST_NAME = 'urn:oid:1.2.840.113994.200.47';
+    public const PREFERRED_MIDDLE_NAME = 'urn:oid:1.2.840.113994.200.48';
+    public const PREFERRED_SURNAME = 'urn:oid:1.2.840.113994.200.49';
+    public const REGISTERED_GIVEN_NAME = 'urn:oid:1.2.840.113994.200.32';
```
lucasmichot, line 16: `REGISTERED_GIVEN_NAME` is not used anywhere

lucasmichot (commented):
Can you check my comments?

src/Saml2/Provider.php
```diff
@@ -108,30 +168,68 @@ protected function getConfig($key = null, $default = null)
...
+            ->getFirstIdpSsoDescriptor()
+            ->getFirstSingleLogoutService($bindingType);
+
+        $logoutResponse = new LogoutResponse();
+        $logoutResponse
+            ->setID(Helper::generateID())
+            ->setInResponseTo($this->messageContext->getMessage()->getID())
+            ->setRelayState($this->messageContext->getMessage()->getRelayState())
+            ->setIssueInstant(new DateTime())
+            ->setDestination($identityProviderLogoutService->getLocation())
```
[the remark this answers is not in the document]
danielztolnai: See my answer above. In case you think this is of concern, I can add extra checks.

src/Saml2/OasisAttributeNameUris.php
danielztolnai: Yes, some of these are not used anywhere. They are standardized URIs and I included them so developers can configure their `attribute_map`s with ease.

Do you think they should not be included? Or should I add some info about these in the readme? (I could also provide a pointer to the other collection (`ClaimTypes`) which includes an other standard as a note on how to configure custom attributes.)

Jellyfrog (approved):
Tested with google saml

src/Saml2/Provider.php
```diff
@@ -255,14 +380,33 @@ protected function getAssertionConsumerServiceRoute(): string
...
+        $methods = [
+            SamlConstants::BINDING_SAML2_HTTP_REDIRECT => 'GET',
+            SamlConstants::BINDING_SAML2_HTTP_POST     => 'POST',
+        ];
+
+        if (!array_key_exists($bindingType, $methods)) {
+            return false;
+        }
+
+        return Arr::has(Route::getRoutes()->getRoutesByMethod(), $methods[$bindingType].'.'.$route);
```
Jellyfrog, line 409: This won't work when you have a route such as;
```
Route::match(['get', 'post'], '{provider}/callback', [SocialiteController::class, 'callback'])->name('callback');
```
(the relevant part being `{provider}`, ie. a "dynamic" url.

I know it was like before also, but maybe we can improve it here.

Something like this should work better;
```php
try {
  return Arr::has(
      app('router')->getRoutes()->match(app('request')->create($route, $methods[$bindingType]))->methods(),
      $methods[$bindingType]
  );
} catch(\Exception $e) {}

return false;
```

Only thing that doesn't work is if you have a "catch all" route.

src/Saml2/README.md
```diff
@@ -103,6 +103,51 @@ Route::post('/auth/callback', function () {
...
+```
+
+#### Single logout
+
+**Warning!** Please note that the SAML2 Single Logout feature is a best effort way of centralized logout. With the current state of affairs it requires special circumstances to work. You have to set your session config `same_site = 'none'` and `secure = true` for it to work which has serious security implications. Please always make sure you understand the risks before using this feature.
+
+You can enable the SingleLogoutService on your Service Provider by adding a GET route where you log the user out and generate the SAML2 logout response:
+```php
+Route::get('/auth/saml2/logout', function () {
+    $response = Socialite::driver('saml2')->logout();
```
ManuelLeiner, line 120:
```suggestion
    $response = Socialite::driver('saml2')->logoutResponse();
```

ManuelLeiner: Is this a typo?

ManuelLeiner:
Tested with Keycloak SAML.

netpok:
We updated the LightSAML package to use the upstream one since the dropbox one does not support Symfony 6 and Laravel 9.

27pchrisl (approved):

27pchrisl:
I'm not sure of the process as I don't have write access - @atymic I think we're good to merge this?

lucasmichot (approved):
GTM @atymic

atymic (approved):

atymic:
Thanks for your work everyone, apologies for the delayed merge. Have a good easter :)

PR #785, opened 2021-12-29
Twitch: Change nickname to use the login value

ekinof:
I changed the nickname value of socialite to use the login value of twitch, if not we have twice the same value for nickname and name


atymic:
This would be a breaking change for existing clients, though I do agree it's correct.
Maybe for the moment, we can add another username property?

atymic (changes requested):
Breaking change, see comment

PR #796, opened 2022-02-07
Remove socialite from TikTok composer.json

dwightwatson:
I ran into some issues with this constraint installing the TikTok provider with Laravel 9.x.

I can see why the dependency was added - it references a Socialite exception. 

However, other SocialiteProviders that also reference Socialite classes omit `laravel/socialite` from their Composer files. TikTok is the only one that references it. This makes it consistent across the board.

### DIFF
--- a/src/TikTok/composer.json
+++ b/src/TikTok/composer.json
@@ -17,7 +17,6 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
-        "laravel/socialite": "~4.0 || ~5.0",
         "socialiteproviders/manager": "~4.0"
     },
     "autoload": {


atymic:
Do we need to change the exception if it's not in this package?

dwightwatson:
Could do - though it wouldn't only apply here as Socialite is referenced directly in a number of providers: https://github.com/SocialiteProviders/Providers/search?q=use+laravel%5Csocialite

atymic (approved):

thinkverse:
Jumping in here real quick.

Think we can safely remove Socialite from all the providers given the manager [requires](https://github.com/SocialiteProviders/Manager/blob/4e63afbd26dc45ff263591de2a0970436a6a0bf9/composer.json#L35) it? It'll be available for each provider via the manager, so I see no reason to explicitly require it **again** in the providers as well.

Just my 2 cents.

PR #799, opened 2022-02-10
Added Bexio Provider

alorian:
Bexio is a service for companies with it's own applications marketplace.
https://www.bexio.com/en-CH/

It would be good having it as a new socialite provider.

### DIFF
--- a/src/Bexio/Provider.php
+++ b/src/Bexio/Provider.php
@@ -53,7 +53,7 @@ class Provider extends AbstractProvider
      */
     protected function mapUserToObject(array $user): User
     {
-        return (new User)->setRaw($user)->map([
+        return (new User())->setRaw($user)->map([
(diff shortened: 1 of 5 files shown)


src/Bexio/Provider.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function mapUserToObject(array $user): User
+    {
+        $fullName = [];
+        if (!empty($user['given_name'])) {
+            $fullName[] = $user['given_name'];
+        }
+        if (!empty($user['given_name'])) {
```
thinkverse, line 59: This if statement is checking if the wrong key is empty.

```suggestion
        if (!empty($user['family_name'])) {
```

lucasmichot (approved):
GTM @atymic

atymic (approved):

PR #801, opened 2022-02-14
Bump JWT dep for eveonline provider

billypoke:
The algorithm argument for the decode function has been removed as of [6.0.0](https://github.com/firebase/php-jwt/releases/tag/v6.0.0)

The [keyset](https://login.eveonline.com/oauth/jwks) contains a key that satisfies the [algorithm](https://github.com/firebase/php-jwt/blob/main/src/JWK.php#L74) for parsing the keys

### DIFF
--- a/src/Eveonline/composer.json
+++ b/src/Eveonline/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
-        "firebase/php-jwt": "^5.2",
+        "firebase/php-jwt": "^6.0",
         "socialiteproviders/manager": "^4.0"
     },
     "autoload": {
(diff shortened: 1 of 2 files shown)


atymic (approved):

src/Eveonline/composer.json
```diff
@@ -18,7 +18,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
-        "firebase/php-jwt": "^5.2",
+        "firebase/php-jwt": "^6.0",
```
lucasmichot, line 21: @atymic why not ^5.2 || ^6.0 ?

lucasmichot: cc @billypoke 
cf. https://github.com/SocialiteProviders/Providers/pull/803

billypoke: See #807

PR #802, opened 2022-02-15
Added client access token method for Okta

clink-aaron:
Okta is heavily used for internal apps that may be authenticating to
each other in scheduled processes where no user token is involved.

This additional method is a QOL for obtaining a client token to
authenticate in these situations.

For additional context, please see the official okta documentation:
https://developer.okta.com/docs/guides/implement-grant-type/clientcreds/main/#about-the-client-credentials-grant

### DIFF
--- a/src/Okta/Provider.php
+++ b/src/Okta/Provider.php
@@ -108,7 +108,8 @@ class Provider extends AbstractProvider
     /**
      * Get the client access token response.
      *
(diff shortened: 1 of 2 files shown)


lucasmichot (approved):
GTM @atymic

atymic (approved):

PR #803, opened 2022-02-16
Bump firebase/php-jwt to v6.0 for apple provider

boris-glumpler:
I am currently using v6 in a project, so can't install the `apple` provider atm. This change seems to have recently been made for the `eveonline` provider already, btw: https://github.com/SocialiteProviders/Providers/pull/801


src/Apple/composer.json
```diff
@@ -34,7 +34,7 @@
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
         "ext-openssl": "*",
-        "firebase/php-jwt": "^5.2",
+        "firebase/php-jwt": "^6.0",
```
lucasmichot, line 37:
```suggestion
        "firebase/php-jwt": "^5.2 || ^6.0",
```

lucasmichot (changes requested):
Thanks for the PR @boris-glumpler 

I am afraid this would introduce a BC
Can you check my review ?

src/Apple/composer.json
boris-glumpler: Sure, that's preferable, but there wasn't much sense in it seeing that `eveonline` only requires v6.0.

atymic:
@boris-glumpler @lucasmichot is this good to go now?

boris-glumpler:
I made the changes, so it's up to @lucasmichot now, I guess.

boris-glumpler:
We have decided to not implement the Apple provider after all and it seems there isn't really much enthusiasm for this PR anyways, so I am closing this now from my end. It is an easy change anyways, so can be resurrected easy enough ;)

Issue #813, opened 2022-03-05
EveOnlineProvider for `"firebase/php-jwt": "^5.2"` broken


herpaderpaldent:
The latest commit to the EveOnline Provider breaks existing installations using `"firebase/php-jwt": "^5.2"`

In order to use `JWT::decode()` with firebase/php-jwt v5.2 one needs to set the used algorithm as it was done prior to:
https://github.com/SocialiteProviders/Providers/commit/6faaad86d37bce45b7856f714a29b692f8497df4#diff-ee2dbdcc9514c641873a948c269b54572040cd0b8d80dac3de71f7381ab9cf08L86 
```php
 $decodedArray = (array) JWT::decode($jwt, JWK::parseKeySet($responseJwksInfo), ['RS256']);
 ``` 
Else this error is being thrown:
https://github.com/firebase/php-jwt/blob/v5.5.1/src/JWT.php#L121
```php
if (empty($algorithm)) {
            // Use deprecated "allowed_algs" to determine if the algorithm is supported.
            // This opens up the possibility of an attack in some implementations.
            // @see https://github.com/firebase/php-jwt/issues/351
            if (!\in_array($header->alg, $allowed_algs)) {
                throw new UnexpectedValueException('Algorithm not allowed');
            }
        }
```

However the firebase/php-jwt v6.0 does not have the third argument anymore https://github.com/firebase/php-jwt/blob/v6.0.0/src/JWT.php#L85

I suggest reverting: https://github.com/SocialiteProviders/Providers/commit/9e494cbfb615866bd6aac4e469fdc007158417f8

atymic:
@herpaderpaldent 

Can you please PR your suggested fix. Can we just re-drop 5.x support, and anyone using jwt 6.x will be able to use the new version?

herpaderpaldent:
Hey @atymic 

i have used your branch to create the PR and add additional instructions on how to delete the current *4.1.0* tag from github and packagist. Revoming this tag is vital.

https://github.com/SocialiteProviders/Providers/pull/815

Grasume:
@atymic Yeah can you push off the #815 branch

herpaderpaldent:
Has been resolved.

PR #815, opened 2022-03-06
fix eve jwt issue

herpaderpaldent:
As discussed in https://github.com/SocialiteProviders/Providers/issues/813

This PR reverts the following PR: "Re-add firebase/php-jwt v5.2 to composer for eveonline provider (#807)"
This reverts commit 9e494cbfb615866bd6aac4e469fdc007158417f8.

### ⚠️ Attention:
Besides merging this PR and tagging you **must** delete the previous 4.1.0 tag not only on github but on packagist too:
https://github.com/SocialiteProviders/Eveonline/tags
image
image

then you can re-tag which should then contain the merged commit/pr with 4.1.0.

If you do not remove the old *4.1.0* tag, composer will still download the broken version from packagist for anyone using the v5 firebase/php-jwt library

### DIFF
--- a/src/Eveonline/composer.json
+++ b/src/Eveonline/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
-        "firebase/php-jwt": "^5.2 || ^6.0",
+        "firebase/php-jwt": "^6.0",
         "socialiteproviders/manager": "^4.0"
     },
     "autoload": {


atymic (approved):

atymic:
Sorry for the delay, been on holiday.

In terms of deleting the tag, I guess that will cause an error for anyone installing, potentially including those who aren't in a broken state.

herpaderpaldent:
@atymic no need to be sorry. I hope you enjoyed your vaccation?

I have previously deleted tags alike from my packages. However it never resulted in DOWNGRADE (see row 6) .From my understanding this is what should happen, after you delete the v4.1.0 tag and version from packagist:

| socialiteproviders/eveonline Version Before | firebase/php-jwt Version  | socialiteproviders/eveonline Version After |
|---------------------------------------------|---------------------------|--------------------------------------------|
| none                                        | ^5.2                      | 4.0.0                                      |
| none                                        | ^6.0                      | 4.2.0                                      |
| 4.0.0                                       | ^5.2                      | 4.0.0                                      |
| 4.0.0                                       | ^6.0                      | 4.2.0                                      |
| **4.1.0**                                   | **^5.2**                  | **4.0.0**                                  |
| 4.1.0                                       | ^6.0                      | 4.2.0                                      |

I can setup an example project and test if in a similar case composer/packagist will serve as expected the downgrade? Would that help? Although i strongly expect composer as descriped above.

herpaderpaldent:
I just tried it with a minimal composer file and ideed it downgrades the version

i started with v4.1.0 and firebase v5.2 and changed the composer.json acoordingly:
image

this results in a downgrade as the version 4.0.0 is the only ledgit combination.

image

atymic:
Done now :)

herpaderpaldent:
Seems to have worked perfectly. Thank you 👍

PR #816, opened 2022-03-09
okta revoke and introspect functionality

clink-aaron:
See documentation here:

- https://developer.okta.com/docs/guides/revoke-tokens/main/
- https://developer.okta.com/docs/reference/api/oidc/#revoke
- https://developer.okta.com/docs/reference/api/oidc/#introspect

### DIFF
--- a/src/Okta/Provider.php
+++ b/src/Okta/Provider.php
@@ -175,4 +175,44 @@ class Provider extends AbstractProvider
 
         return "$url?$params";
     }
+
+    /**
+     * @param string $token
+     * @param string $hint
+     *
+     * @return \Psr\Http\Message\ResponseInterface
+     */
+    public function revokeToken($token, $hint = 'access_token')
+    {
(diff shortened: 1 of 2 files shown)


src/Okta/Provider.php
```diff
@@ -175,4 +175,45 @@ public function getLogoutUrl(string $idToken, string $redirectUri = null, string
...
         return "$url?$params";
     }
+
+    /**
+     * @param string $token
+     * @param string $hint
+     *
+     * @return \Psr\Http\Message\ResponseInterface
+     */
+    public function revokeToken($token, $hint = 'access_token')
```
atymic, line 185: can we have types please?

atymic: (and below)

atymic (changes requested):

src/Okta/Provider.php
clink-aaron: @atymic fixed. Should probably add a check for this in the linter at some point.

Issue #818, opened 2022-03-15
AzureADB2C: call different policies


lsmith77:
I managed to get the integration working for login (policy `B2C_1_signupsignin1`), but now I also want to call other policies, like for example updating the profile (policy `B2C_1_profile_mgmt1`).

However, it seems like `SocialiteProviders\AzureADB2C\Provider` can only handle one policy.

I tried to just define another service via
```
    'azureadb2c' => [
        'client_id' => env('AADB2C_CLIENT_ID'),
        'client_secret' => env('AADB2C_CLIENT_SECRET'),
        'redirect' => '/oauth/azureadb2c/callback',
        'domain' => env('AADB2C_DOMAIN'),
        'policy' => env('AADB2C_POLICY'),
    ],

    'azureadb2c_profile' => [
        'client_id' => env('AADB2C_CLIENT_ID'),
        'client_secret' => env('AADB2C_CLIENT_SECRET'),
        'redirect' => '/oauth/azureadb2c_profile/callback',
        'domain' => env('AADB2C_DOMAIN'),
        'policy' => env('AADB2C_POLICY_PROFILE'),
    ],
```

But that obviously doesn't work since there is no driver for `azureadb2c_profile`.

Am I overlooking something here?

PR #819, opened 2022-03-15
AzureADB2C: make it possible to define multiple policies

lsmith77:
fixes https://github.com/SocialiteProviders/Providers/issues/818

This all feels quite hacky, especially overriding the `redirectUrl` but I didn't manage to figure out another way.

Note I am using SocialStream, so I had to override quite a few things there as well, to pass the policy as a parameter to the `Provider`

```
Socialite::driver($provider)
  ->with(['policy' => $policy])
  ->redirect();
```

I hope I sufficiently taken care of backwards compatibility.

### DIFF
--- a/src/AzureADB2C/Provider.php
+++ b/src/AzureADB2C/Provider.php
@@ -26,11 +26,21 @@ class Provider extends AbstractProvider
         'openid',
     ];
 
+    /**
+     * Get the policy.
+     *
+     * @return string
+     */
     private function getPolicy()
     {
         return $this->parameters['policy'] ?? 'login';
     }
 
+    /**
+     * Get the B2C policy.
+     *
+     * @return mixed
+     */
     private function getB2CPolicy()
     {
         $policy = $this->getConfig('policy');
@@ -42,6 +52,11 @@ class Provider extends AbstractProvider
         return $policy;
     }
 
+    /**
+     * Set the redirect URL.
+     *
+     * @return void
+     */
     private function setRedirectUrl()
     {
         $redirectTemplate = $this->getConfig('redirect_template');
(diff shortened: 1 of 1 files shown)


src/AzureADB2C/Provider.php
```diff
@@ -148,7 +187,6 @@ protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
             'id'   => $user['sub'],
-            'name' => $user['name'],
```
lsmith77, line 151: this should be left to the user landcode to determine what should be mapped

lsmith77:
@fujie what do you think about this change?

fujie:
@lsmith77 Let me understand what you're trying to do. Is it correct that you want to use multiple policies for the same application by overriding policy and redirect_uri when calling the driver, and you're proposing to change default claim mapping on Provider.php to map only sub claim to override the claims mapping for every policies?
If I understand correctly, I think it is reasonable change.

lsmith77:
yeah exactly!

lsmith77:
@lucasmichot given that @fujie agrees with the change, are you willing to merge this PR?

lucasmichot:
> @lucasmichot given that @fujie agrees with the change, are you willing to merge this PR?

We typically prefer having a double review @lsmith77 , as soon as an other member will review 🟢  this PR, it will be merged.

lucasmichot (approved):
Thanks a lot for the PR @lsmith77

lsmith77:
ok thank you. Will see if I can get another person to review.

f.e. I have now also implemented a way to get an `access_token`. but not sure if this is sufficiently generic and definately goes beyond the scope of this PR. just leaving it here for now.

```
    public function getAccessTokenResponse($code)
    {
        $fields = $this->getTokenFields($code);
        $fields['scope'] = implode(' ', $this->getScopes());

            RequestOptions::FORM_PARAMS => $fields,
        ]);

    }
```

atymic (approved):
LGTM 👍

PR #821, opened 2022-03-23
Azure Provider - Fix when login with Guest account types

moisish:
Hello,

Azure is a weird thing and when you are having guest account types on your organisation the User Principal Name has the format of ``` my-email_gmail.com#EXT#@myorganisation.com```

The change I suggest is to use the 'mail' field on the User object for the email field to make sure the correct email address is returned even for guest accounts.

I have tested this on a private application I am working and it works for both member and guest account types.


moisish:
@lucasmichot Can you review this please?

atymic (changes requested):
I guess this would be considered a breaking change, since any accounts logged in will no longer be able to log in?

Maybe instead we could add a flag and a way of getting the underlying email?

moisish:
@atymic Based on my testing the change does not affect existing accounts and logins because the mail field will return the same email address with the principal name for normal azure accounts. 

I am happy if someone else can verify this.

If you still thing this may be a breaking change otherwise I update my PR by having some replication in the mean time until a major version update.

```
            'email'         => $user['userPrincipalName'],
            'mail' => $user['mail']
            'principalName' => $user['userPrincipalName'],
```

moisish:
@atymic any news on this?

atymic:
Hey @moisish 

Yes, can we add a new field rather than changing the existing one unless there's docs you can link to :)

atymic:
cc @moisish can you add the new field?

atymic:
ping @moisish

moisish:
@atymic Sorry for the late response. I have added the new field

atymic (approved):

PR #822, opened 2022-03-25
stateless fixed default

condor-bird:
The current stateless property is set to true. This prevents the state option from being used, while setting stateless to false allows it to be used and throws \InvalidStateException as it should work.

### DIFF
--- a/src/VKontakte/Provider.php
+++ b/src/VKontakte/Provider.php
@@ -20,7 +20,7 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
-    protected $stateless = true;
+    protected $stateless = false;
 
     /**
      * {@inheritdoc}


lucasmichot:
Thanks @condor-bird , can you point out to the API documentation ?

condor-bird:
> Thanks @condor-bird , can you point out to the API documentation ?

Hi! If I understand you, then to the VK API documentation? I will describe in a little more detail, because I did not understand a little. I fixed this error using the example of the standard Socialite documentation: https://socialiteproviders.com/VKontakte/#installation-basic-usage

```
Socialite::driver('vkontakte')->redirect();
```

Current change:

```
protected $stateless = false;
```
Should not affect the health of current users of the package. On the contrary, it will help prevent request spoofing without state, in which case access to someone else's account can be obtained where VK authorization is used without state. Those who don't want to use state (though why) can still use ->stateless().
Thus, this will now work fine state as it currently works in Google, Facebook Auth and other similar ones.

condor-bird:
Problem actual

atymic (approved):

atymic:
Since this is a BC break i'll tag a new major version

PR #826, opened 2022-04-04
Add Starling Provider

lukecurtis93:
This PR allows for Starling oAuth2 flow to be utilised. 
- Starling have particular endpoints for sandbox and have therefore followed similar spec in QuickBooks provider for determining which endpoint to utilise
- Also, they do not return the unique user ID as part of their identity call and it must be deduced by the token call. For now I have merged the two results and am in discussions with their technical team to see if there is a more accessible way of obtaining this information.

___
Website: https://www.starlingbank.com/
___
Docs: https://developer.starlingbank.com/docs


lucasmichot (approved):
Thanks for the PR @lukecurtis93

atymic (approved):

PR #829, opened 2022-04-12
Add Toyhou.se provider

sarahjting:
Provider for https://toyhou.se/
Developer documentation: https://toyhou.se/~developer/oauth2

Thank you! :)

### DIFF
--- a/src/Toyhouse/Provider.php
+++ b/src/Toyhouse/Provider.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace SocialiteProviders\Toyhouse;
+
+use GuzzleHttp\RequestOptions;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    public const IDENTIFIER = 'TOYHOUSE';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = [''];
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state)
+    {
+        return $this->buildAuthUrlFromBase('https://toyhou.se/~oauth/authorize', $state);
+    }
+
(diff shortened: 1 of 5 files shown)


src/Toyhouse/Provider.php
```diff
@@ -0,0 +1,64 @@
...
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    public const IDENTIFIER = 'TOYHOUSE';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = [''];
```
atymic, line 19: Don't need this, default is empty

src/Toyhouse/composer.json
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Sarah Ting",
+            "email": "sarah.j.ting@gmail.com"
+        }
+    ],
+    "require": {
+        "php": "^7.2 || ^8.0",
```
atymic, line 19:
```suggestion
        "php": "^7.4 || ^8.0",
```

atymic (changes requested):

sarahjting:
Thank you for review! Have updated 🙇 

It's an art gallery site, similar to deviantArt but supports trading/selling art collections! It's relatively small, but I didn't see any requirements for what sort of providers you're accepting so thought I'd see if I could get it added hehe

src/Toyhouse/README.md
```diff
@@ -0,0 +1,42 @@
...
+];
+```
+
+### Usage
+
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
+
+```php
+return Socialite::driver('toyhouse')->redirect();
+```
```
atymic, line 42: Can you add the returned fields here? See other providers for example on website.

Thanks :)

atymic (changes requested):
One minor but good to go otherwise

sarahjting:
Added! Thanks again ☺️

atymic (approved):

Issue #824, opened 2022-03-30
getAvatar broken for Microsoft provider


has-n:
```Client error: `GET https://graph.microsoft.com/v1.0/me/photo/$value` resulted in a `404 Not Found` response: {"error":{"code":"UnknownError","message":"{\r\n \"errorCode\": \"ErrorProfileNotFound\",\r\n \"message\": \"Exception (truncated...)```

404 is returned if the user has not assigned a profile picture. The endpoint can be tested before returning null.

atymic:
Feel free to make a PR wrapping it in a try/catch :)

PR #830, opened 2022-04-13
Turn off http errors when fetching Microsoft avatars

elite123:
Fixes https://github.com/SocialiteProviders/Providers/issues/824 where a 404 error bubbles up if the avatar doesn't exist

### DIFF
--- a/src/Microsoft/MicrosoftUser.php
+++ b/src/Microsoft/MicrosoftUser.php
@@ -14,6 +14,7 @@ class MicrosoftUser extends User
     public function getAvatar()
     {
         $client = new Client();
+        
         try {
             $response = $client->get(
                 'https://graph.microsoft.com/v1.0/me/photo/$value',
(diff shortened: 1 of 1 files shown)


atymic (commented):
Should we not use a try/catch and return null?

elite123:
@atymic I have updated to catch \GuzzleHttp\Exception\ClientException

src/Microsoft/MicrosoftUser.php
```diff
@@ -14,16 +14,21 @@ class MicrosoftUser extends User
...
+                [
+                    RequestOptions::HEADERS => [
+                        'Accept'        => 'image/*',
+                        'Authorization' => 'Bearer '.$this->token,
+                    ],
+                ]
+            );
+
+            return (new MicrosoftAvatar())->setResponse($response);
+        } catch (\GuzzleHttp\Exception\ClientException $e) {
```
lucasmichot, line 30: Can you import this `ClientException` ?

lucasmichot (changes requested):

src/Microsoft/MicrosoftUser.php
elite123: @lucasmichot Updated to import `ClientException`

atymic (approved):

atymic (approved):

Issue #835, opened 2022-04-18
Discord Package Needs PKCE Enabled


ferrari308guy:
While trying to retrieve the user information using the Laravel Socialite Discord package, I receive the following error:

> Client error: `POST https://discord.com/api/oauth2/token` resulted in a `400 Bad Request` response: {"error": "invalid_request", "error_description": "Missing \"code_verifier\""}

I tracked this issue down to the fact that this package needs to have PCKE enabled so that the `code_verifier` is passed along with the request for user information.

The fix is a one-line change in [this file](https://github.com/SocialiteProviders/Discord/blob/master/Provider.php). The following definition needs to be added:

> /**
     * Indicates if PKCE should be used.
     *
     * @var bool
     */
    protected $usesPKCE = true;

I saw that repo is marked as Read-Only, so I'm posting here because I'm not sure what the PR process is to get that fixed.

atymic:
Just PR against this repo to the discord folder :)

ferrari308guy:
Thanks. [Here's](https://github.com/SocialiteProviders/Providers/pull/836) the PR.

greatsaltlake:
For what it's worth, I was getting the same response back from Discord which led me to believe PCKE being disabled was the issue. After a bit more debugging, it turned out I had a couple of config errors. My client secret was empty due to using the wrong ENV name, and my session domain was wrong due to using ngrok for local testing, so none of the session values were getting set. So if you're getting this error it's possible you have a different issue.

ferrari308guy:
I looked into this and I also had a problem with the client secret ENV variable not being configured properly. After fixing that issue, this works without PKCE being enabled. I'll close the PR since it's not needed.

PR #836, opened 2022-04-19
Enable PKCE for Discord

ferrari308guy:


atymic:
I had a look at the docs and didn't see any mention of PKCE
https://discord.com/developers/docs/topics/oauth2

Can you link the docs please?

ferrari308guy:
> I had a look at the docs and didn't see any mention of PKCE https://discord.com/developers/docs/topics/oauth2
> 
> Can you link the docs please?

Yes, I couldn't find any mention of PKCE in the docs, either.

I tried to implement the Discord package as-is via Laravel Socialite. When trying to access the user property on the callback function via the following:

`$user = Socialite::driver('discord')->user();`

I received a 400 error from the server with the message that I was missing a `code_verifier` parameter. I did some digging and it appears this parameter is tied to PKCE, which led me to conclude that PKCE needs to be enabled for Discord. After enabling PKCE via the change I propose in this PR, I was able to access the user parameter without issue.

atymic:
See this, seems it could be something else @ferrari308guy 
https://github.com/SocialiteProviders/Providers/issues/835?

ferrari308guy:
I made a comment in #835. I had a configuration issue on my end that was the source of the problem. I'll close this PR.

Issue #820, opened 2022-03-16
Repo Subtree Split is broken


atymic:
We're ending up with only a single commit in each repo, previously this had the proper history. Investigate.

atymic:
@lucasmichot not sure if this was a side effect of upgrading some of our actions workflows at some point. Previously the subtree split has all the commit history for that folder, now it seems we only get the single latest commit, even if it's for a different provider. Any ideas?

lucasmichot:
Thanks for opening the ticket @atymic, I have investigated and could see any similar issue during the upgrade for v2 to v3.
I have nevertheless created a PR worth testing: #838

PR #838, opened 2022-04-20
Do not limit the number of commits to fetch

lucasmichot:
As per #820, the upgrade of https://github.com/actions/checkout v3 has broken the subtree split.
This seems related to the depth (or history) this change allows to get all history for all branches and tags.

### DIFF
--- a/.github/workflows/subtree-pr-close.yml
+++ b/.github/workflows/subtree-pr-close.yml
@@ -11,6 +11,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


atymic (approved):
Thanks, hopefully fixes it 🤞

atymic:
Hmm, didn't fix it. I'll try locally and see if I can repro it.

lucasmichot:
Thanks for testing @atymic 
Let me know how this work, in the worst case I guess we could revert back to https://github.com/actions/checkout v2

lucasmichot:
@atymic , see #840 that rollbacks `actions/checkout` to `v2`.

PR #841, opened 2022-04-26
Update Cognito Provider

JamesSiebert:
Removed Cognito provider fields that can cause errors.
Errors were occurring when the Cognito user pool doesn't have these same fields defined.

### DIFF
--- a/src/Cognito/Provider.php
+++ b/src/Cognito/Provider.php
@@ -91,6 +91,7 @@ class Provider extends AbstractProvider
             'nickname' => $user['nickname'] ?? null,
             'name'     => trim(Arr::get($user, 'given_name', '').' '.Arr::get($user, 'family_name', '')),
             'email'    => $user['email'] ?? null,
+            'avatar'   => null,
         ]);
     }
 
--- a/src/Cognito/Provider.php
+++ b/src/Cognito/Provider.php
@@ -2,6 +2,7 @@
 
 namespace SocialiteProviders\Cognito;
 
+use Illuminate\Support\Arr;
 use GuzzleHttp\RequestOptions;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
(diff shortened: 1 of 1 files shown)


src/Cognito/Provider.php
```diff
@@ -86,13 +85,7 @@ protected function getUserByToken($token): array
      */
     protected function mapUserToObject(array $user)
     {
-        return (new User())->setRaw($user)->map([
-            'id'       => $user['sub'],
-            'nickname' => $user['nickname'],
-            'name'     => trim(Arr::get($user, 'given_name', '').' '.Arr::get($user, 'family_name', '')),
-            'email'    => $user['email'],
```
atymic, line 93: Why not default these to null instead of removing them completely? this will be a BC break.

atymic (changes requested):

src/Cognito/Provider.php
JamesSiebert: Good call, I was thinking that having all of the fields in one user array may be cleaner but I think this is ok, nulls added. Thanks :)

src/Cognito/Provider.php
```diff
@@ -87,11 +87,10 @@ protected function getUserByToken($token): array
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'       => $user['sub'],
-            'nickname' => $user['nickname'],
+            'id'       => $user['sub'] ?? null,
+            'nickname' => $user['nickname'] ?? null,
             'name'     => trim(Arr::get($user, 'given_name', '').' '.Arr::get($user, 'family_name', '')),
-            'email'    => $user['email'],
-            'avatar'   => null,
```
lucasmichot, line 94: Can you re-add the avatar field ?

lucasmichot (changes requested):

src/Cognito/Provider.php
JamesSiebert: Sorry forgot to add back in, done :)

lucasmichot (approved):

JamesSiebert:
Hey :) is there anything else I need to do here?

PR #842, opened 2022-04-27
Fix typo in Saml2 readme

tontonsb:
This appears to be a typo as the `logout` method doesn't exist.

I haven't gotten the feature to work yet so maybe @danielztolnai can confirm if the fix that I propose is correct.


danielztolnai:
You are right, this is just a typo in the readme, the change you propose fixes it.

atymic (approved):

PR #846, opened 2022-05-12
apple: Support for newer versions

lukasjuhas:
I have a project using Laravel, and I've been trying to upgrade to Laravel 9, however, this package manager was locked at 4.0. Newer version 4.1 has added support for laravel 9 which I'm after.

### DIFF
--- a/src/Apple/composer.json
+++ b/src/Apple/composer.json
@@ -36,7 +36,7 @@
         "ext-openssl": "*",
         "firebase/php-jwt": "^5.2",
         "lcobucci/jwt": "^4.0",
-        "socialiteproviders/manager": "~4.0"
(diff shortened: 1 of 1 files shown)


src/Apple/composer.json
```diff
@@ -36,7 +36,7 @@
         "ext-openssl": "*",
         "firebase/php-jwt": "^5.2",
         "lcobucci/jwt": "^4.0",
-        "socialiteproviders/manager": "~4.0"
+        "socialiteproviders/manager": "^4.0"
```
atymic, line 39: `~4.0` allows 4.1. Can you post the output of composer update, something else is causing the conflict.

atymic (changes requested):

PR #847, opened 2022-05-17
fix Mailru mapUserToObject id attribute

Cellard:


src/Mailru/Provider.php
```diff
@@ -53,7 +53,7 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'       => $user['email'],
+            'id'       => $user['id'],
```
atymic, line 56: this would be a breaking change i guess?

Cellard: Likely. But it is anyway broken for now (

Email is not an identifier. And MailRu provides an identifier.

atymic (approved):

Cellard:
@atymic I forget to fix README for the provider. Should I make new PR or maybe you fix docs yourself?

### Returned User fields

- ``id`` ~~(same as ``email``)~~
- ``nickname``
- ``name``
- ``email``
- ``avatar``

atymic:
Fixed :)

PR #852, opened 2022-05-21
upgrade lcobucci/jwt to 4.1.5

Alsharie:
upgrade jwt package to support "LooseValidAt"

image


Alsharie:
upgrade  lcobucci/jw from _4.0_ to _4.1.5_ in **apple provider**  in order to use **LooseValidAt** constraint

atymic (approved):

PR #861, opened 2022-07-15
added a request for revoke Apple token

bearn666:
Based on the last updates from Apple, where we should add to the user's ability to delete an account, we also should revoke his tokens [(Apple doc).](https://developer.apple.com/documentation/sign_in_with_apple/revoke_tokens) So I think will be good to have this ability out of the box, as the "Okta" provider has. [(Okta)](https://socialiteproviders.com/Okta/#installation-basic-usage)

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -241,4 +241,30 @@ class Provider extends AbstractProvider
 
         return json_decode($value, true);
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


atymic (approved):

atymic:
https://github.com/SocialiteProviders/Apple/releases/tag/5.3.0

Issue #860, opened 2022-06-24
SAML2 XML Configuration Digest Error


robertdeboer:
Hello all,

Thank you for this wonderful library. I have come across what I think is a bug when digesting a valid SAML 2 XML configuration generated by a Identity Provider for use in a service provider application.

Back story:
This is in used on a Laravel 8 application, using Laravel Socialite with Socialite Providers SAML2
```
"laravel/framework": "^8.12",
"laravel/socialite": "^5.5",
"socialiteproviders/saml2": "dev-master"
```
SAML2 is on dev-master simply so that I could utilize it's latest updates, which is switching over to litesaml/lightsaml from lightsaml/lightsaml. I think this issue is present in either master or a tag branch release.

This application is a simple Service Provider, with the Identity Provider being managed by a third party enterprise system. They provided the XML below so I could ingest and get my needed configuration. I have ran it through https://www.samltool.com/validate_xml.php using the Metadata XSD schema and it passes.

My SAML 2 XML Config
```xml
<EntitiesDescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" ID="ABC123ABC123ABC123ABC123ABC123ABC123ABC123AB">
    <ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
        <ds:SignedInfo>
            <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
            <ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
            <ds:Reference URI="#DEF456DEF456DEF456DEF456DEF456DEF456DEF456DE">
                <ds:Transforms>
                    <ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
                    <ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
                </ds:Transforms>
                <ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
                <ds:DigestValue>xxxxxxxxxxxxxxxxxxxxxxxxxxxx</ds:DigestValue>
            </ds:Reference>
        </ds:SignedInfo>
        <ds:SignatureValue>
[...]

PR #862, opened 2022-07-16
Handle multiple entity descriptors in IdP metadata, by choosing the first one.

27pchrisl:
If the metadata provided by the IDP had multiple entity descriptors then this was not properly handled.

Fixes the issue described in #860. Props @robertdeboer.


atymic:
Thanks!

PR #869, opened 2022-07-27
Added Socialite provider for Authentik

rf152:
Added a socialite provider for [Authentik](https://goauthentik.io/)

Base_URL is extracted to allow customisation to a specific Authentik instance

### DIFF
--- a/src/Authentik/Provider.php
+++ b/src/Authentik/Provider.php
@@ -2,8 +2,8 @@
 
 namespace SocialiteProviders\Authentik;
 
-use InvalidArgumentException;
 use GuzzleHttp\RequestOptions;
+use InvalidArgumentException;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
 
--- a/src/Authentik/Provider.php
+++ b/src/Authentik/Provider.php
@@ -15,7 +15,7 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
-    protected $scopes = ['openid goauthentik.io/api profile'];
+    protected $scopes = ['openid goauthentik.io/api profile email'];
 
     /**
      * {@inheritdoc}
@@ -66,8 +66,11 @@ class Provider extends AbstractProvider
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'email'    => $user['sub'],
+            'email'    => $user['email'],
             'name'     => $user['name'],
+            'nickname' => $user['nickname'],
+            'groups'   => $user['groups'],
+            'email_verified' => $user['email_verified'],
         ]);
     }
 
--- a/src/Authentik/Provider.php
+++ b/src/Authentik/Provider.php
@@ -3,6 +3,7 @@
 namespace SocialiteProviders\Authentik;
 
 use InvalidArgumentException;
+use GuzzleHttp\RequestOptions;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
 
@@ -28,7 +29,7 @@ class Provider extends AbstractProvider
 
     protected function getBaseUrl()
     {
-        $baseurl = $this->getConfgi('base_url');
+        $baseurl = $this->getConfig('base_url');
         if ($baseurl === null) {
             throw new InvalidArgumentException('Missing base_url');
         }
@@ -56,11 +57,14 @@ class Provider extends AbstractProvider
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get($this->getBaseUrl().'/application/o/userinfo/', [
-            'headers' => [
-                'Authorization' => 'Bearer '.$token,
-            ],
-        ]);
+        $response = $this->getHttpClient()->get(
+            $this->getBaseUrl().'/application/o/userinfo/',
+            [
+                RequestOptions::HEADERS => [
+                    'Authorization' => 'Bearer '.$token,
+                ],
+            ]
+        );
 
         return json_decode((string) $response->getBody(), true);
     }
@@ -79,7 +83,7 @@ class Provider extends AbstractProvider
             'preferred_username' => $user['preferred_username'] ?? '',
             'nickname'           => $user['nickname'] ?? '',
             'groups'             => $user['groups'] ?? '',
-            'id'                => $user['sub'],
+            'id'                 => $user['sub'],
         ]);
     }
 }
--- a/src/Authentik/Provider.php
+++ b/src/Authentik/Provider.php
@@ -76,7 +76,7 @@ class Provider extends AbstractProvider
     protected function getTokenFields($code)
     {
         return array_merge(parent::getTokenFields($code), [
-            'grant_type' => 'authorization_code'
+            'grant_type' => 'authorization_code',
         ]);
     }
 }
--- a/src/Authentik/Provider.php
+++ b/src/Authentik/Provider.php
@@ -76,14 +76,14 @@ class Provider extends AbstractProvider
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'email'              => $user['email'] ?? '',
+            'email'              => $user['email'] ?? null,
             'email_verified'     => $user['email_verified'] ?? null,
-            'name'               => $user['name'] ?? '',
-            'given_name'         => $user['given_name'] ?? '',
-            'family_name'        => $user['family_name'] ?? '',
-            'preferred_username' => $user['preferred_username'] ?? '',
-            'nickname'           => $user['nickname'] ?? '',
-            'groups'             => $user['groups'] ?? '',
+            'name'               => $user['name'] ?? null,
+            'given_name'         => $user['given_name'] ?? null,
+            'family_name'        => $user['family_name'] ?? null,
+            'preferred_username' => $user['preferred_username'] ?? null,
+            'nickname'           => $user['nickname'] ?? null,
+            'groups'             => $user['groups'] ?? null,
             'id'                 => $user['sub'],
         ]);
     }
--- a/src/Authentik/Provider.php
+++ b/src/Authentik/Provider.php
@@ -15,7 +15,7 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
-    protected $scopes = ['openid goauthentik.io/api'];
+    protected $scopes = ['openid goauthentik.io/api profile'];
 
     /**
      * {@inheritdoc}
@@ -67,6 +67,7 @@ class Provider extends AbstractProvider
     {
         return (new User())->setRaw($user)->map([
             'email'    => $user['sub'],
+            'name'     => $user['name'],
         ]);
     }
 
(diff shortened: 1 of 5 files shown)


AlexZorzi:
Great! bump PR
We really need this 👍

src/Authentik/README.md
```diff
@@ -0,0 +1,43 @@
...
+];
+```
+
+### Usage
+
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
+
+```php
+return Socialite::driver('authentik')->redirect();
+```
```
atymic, line 43: Can we add returned fields, or at least document how to access the non standard fields?

src/Authentik/composer.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "socialiteproviders/authentik",
+    "description": "Authentik OAuth2 Provider for Laravel Socialite",
+    "license": "MIT",
+    "authors": [{
+        "name": "rf152",
+        "email": "git@rf152.co.uk"
+    }],
+    "require": {
+        "php": "^7.2",
```
atymic, line 10:
```suggestion
        "php": "^8.0",
```

src/Authentik/README.md
```diff
@@ -0,0 +1,43 @@
+# Authentik
+
+```bash
+composer require socialiteproviders/authentik
+```
+
+## Installation & Basic Usage
+
+Please see the [Base Installation Guide](https://socialiteproviders.com/usage/), then follow the provider specific instructions below.
```
atymic, line 9: Can we add a link to the authentik docs?

AlexZorzi: leaving here Oauth2 docs for authentik, just in case https://goauthentik.io/docs/providers/oauth2/ :)

rf152:
I've made changes to the README and composer.json, and a couple of changes to the code itself to return more than just the email address.

AlexZorzi:
> I've made changes to the README and composer.json, and a couple of changes to the code itself to return more than just the email address.

Do you retrive all the available information? On authentik it says openid profile and email are the requestable values

rf152:
Yep, and in the profile against the default Laravel User object the attributes that match are email and name, so it maps both of those.

AlexZorzi:
> Yep, and in the profile against the default Laravel User object the attributes that match are email and name, so it maps both of those.

i would also include all the other available information that could be usefull like 'nickname' 'groups' and others

if you add 'email' like so 'protected $scopes = ['openid goauthentik.io/api profile email'];'

you can also get 'email_verified'

Immagine 2022-07-29 173244

AlexZorzi:
still some values missing, refer to this commit as example, also i added a default value in case authentik dosent provide the full access to profile email and openid

https://github.com/AlexZorzi/Providers/commit/927a5c5802e74cf11fe05d409236278b1341a003

rf152:
I've added all the fields that you've mentioned; not sure how it will work if the user model doesn't have those values - I'm guessing it won't throw an error?

AlexZorzi:
> I've added all the fields that you've mentioned; not sure how it will work if the user model doesn't have those values - I'm guessing it won't throw an error?

nice, yes if the key is not present it will default to "" apart for email_verified that as a boolean will default to null

rf152:
> > I've added all the fields that you've mentioned; not sure how it will work if the user model doesn't have those values - I'm guessing it won't throw an error?
> 
> nice, yes if the key is not present it will default to "" apart for email_verified that as a boolean will default to null

I mean the User model within Laravel. I can see that you've set a default for the ones coming in from OAuth.

More that the default User model in Laravel doesn't have half those fields....

AlexZorzi:
I would say that you can always filter out stuff that you don’t need afterwards. I would give the user all the available data coming from authentik anyway 
Maybe @atymic has a better way in mind to be compatible with laravel default User model without loosing data that could be useful

src/Authentik/composer.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "socialiteproviders/authentik",
+    "description": "Authentik OAuth2 Provider for Laravel Socialite",
+    "license": "MIT",
```
lucasmichot, line 4: Can you also add the missing keywords section?

lucasmichot: Also the support section is missing.

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    const IDENTIFIER = 'AUTHENTIK';
```
lucasmichot, line 13:
```suggestion
    public const IDENTIFIER = 'AUTHENTIK';
```

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+     * {@inheritdoc}
+     */
+    public static function additionalConfigKeys()
+    {
+        return ['base_url'];
+    }
+
+    protected function getBaseUrl()
+    {
+        return rtrim($this->getConfig('base_url'), '/');
```
lucasmichot, line 30: Can you raise an InvalidArgumentException if this config key is no set?

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+        return $this->getBaseUrl().'/application/o/token/';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getUserByToken($token)
+    {
+        $response = $this->getHttpClient()->get($this->getBaseUrl().'/application/o/userinfo/', [
+            'headers' => [
```
lucasmichot, line 55: Please use the RequestOptions headers const instead

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+                'Authorization' => 'Bearer '.$token,
+            ],
+        ]);
+
+        return json_decode($response->getBody(), true);
```
lucasmichot, line 60:
```suggestion
        return json_decode((string) $response->getBody(), true);
```

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+            'nickname'           => $user['nickname'] ?? '',
+            'groups'             => $user['groups'] ?? '',
+            'sub'                => $user['sub'] ?? '',
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getTokenFields($code)
```
lucasmichot, line 84: Useless as the parent already sets the grant_type

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+        return json_decode($response->getBody(), true);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function mapUserToObject(array $user)
+    {
+        return (new User())->setRaw($user)->map([
+            'email'              => $user['email'] ?? '',
```
lucasmichot, line 69: See `\Laravel\Socialite\Contracts\User` for the required fields
You need at mminima an email, and id or a nickname to identify the user.

check for example https://github.com/laravel/socialite/blob/5.x/src/Two/GoogleProvider.php#L72-L78

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    const IDENTIFIER = 'AUTHENTIK';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = ['openid goauthentik.io/api profile email'];
```
lucasmichot, line 18: This scope format looks like an error

lucasmichot (changes requested):

lucasmichot (changes requested):
This provider is missing in `monorepo-builder.yml`

Please double check https://socialiteproviders.com/contribute/#creating-a-provider

src/Authentik/Provider.php
AlexZorzi: Am I missing something?
In the part of the code you are referring an email and a nickname are already set

lucasmichot: Both email and nickname can be empty string.
And no user id is provided.

If you don't provide a unique user id, you need to provide some non-empty unique strings somewhere, right ?

src/Authentik/Provider.php
rf152: It's a valid scope format. (tested and working)

src/Authentik/Provider.php
rf152: I've adjusted the returned values to return the sub as an id field. That will never be empty (it's the subject of the authentication)

src/Authentik/Provider.php
rf152: Removed

src/Authentik/Provider.php
rf152: I don't know what you mean by this one - are you able to give an example?

src/Authentik/Provider.php
rf152: done

src/Authentik/Provider.php
rf152: done

src/Authentik/composer.json
rf152: Added

src/Authentik/Provider.php
rf152: done

src/Authentik/Provider.php
lucasmichot: Check other providers

src/Authentik/Provider.php
```diff
@@ -0,0 +1,85 @@
...
+     * {@inheritdoc}
+     */
+    public static function additionalConfigKeys()
+    {
+        return ['base_url'];
+    }
+
+    protected function getBaseUrl()
+    {
+        $baseurl = $this->getConfgi('base_url');
```
lucasmichot, line 31: typo

src/Authentik/Provider.php
rf152: Done

lucasmichot (approved):
Thx for the PR @rf152 
GTM @atymic

src/Authentik/Provider.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return (new User())->setRaw($user)->map([
+            'email'              => $user['email'] ?? '',
+            'email_verified'     => $user['email_verified'] ?? null,
+            'name'               => $user['name'] ?? '',
+            'given_name'         => $user['given_name'] ?? '',
+            'family_name'        => $user['family_name'] ?? '',
+            'preferred_username' => $user['preferred_username'] ?? '',
+            'nickname'           => $user['nickname'] ?? '',
+            'groups'             => $user['groups'] ?? '',
```
atymic, line 86: Is there a reason we default to empty string here, not null?
I think null makes mores sense - the value isn't present

atymic (commented):
Just one minor change

src/Authentik/Provider.php
AlexZorzi: it was my suggestion, but yes, returning null would be better

atymic (approved):

AlexZorzi:
@atymic how long does it take for the changes to be available on composer?

atymic:
https://packagist.org/packages/socialiteproviders/authentik

atymic:
Please also PR to the website repo

PR #872, opened 2022-07-31
Bump to PHP 8

lucasmichot:
PHP 7.4 will be EOL in Nov. 2022.

This PR bumps PHP minimum requirement to `^8.0`.

This change requires a new major release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         validate: [ false ]
         fail-fast: [ false ]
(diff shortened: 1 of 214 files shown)


atymic:
I think we should wait until the EOL before forcing an upgrade, esp. if it's just to change code style tools?

lucasmichot:
> I think we should wait until the EOL before forcing an upgrade, esp. if it's just to change code style tools?

No, it's actually relate to the Pint PR
This is really to only to ensure the package is supported by fairy recent versions of PHP

PR #882, opened 2022-08-06
No need to set the grant_type field again

lucasmichot:
No need to set the `grant_type` field again, it is already set with the value `authorization_code` in the parent method.

---

Requires https://github.com/SocialiteProviders/Providers/pull/899

### DIFF
--- a/src/Acclaim/Provider.php
+++ b/src/Acclaim/Provider.php
@@ -61,6 +61,4 @@ class Provider extends AbstractProvider
             'avatar'   => $user['data']['photo_url'],
         ]);
     }
-
-
 }
--- a/src/Acclaim/Provider.php
+++ b/src/Acclaim/Provider.php
@@ -62,13 +62,5 @@ class Provider extends AbstractProvider
         ]);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function getTokenFields($code)
-    {
-        return array_merge(parent::getTokenFields($code), [
-            'grant_type' => 'authorization_code',
-        ]);
-    }
+
 }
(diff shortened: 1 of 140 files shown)


atymic (changes requested):
https://github.com/SocialiteProviders/Manager/blob/master/composer.json#L35

This allows v4 or v5, meaning if the user has v4, which doesn't have the grant type won't the provider break?
I guess to merge this we'd need to bump the min to v5 on manager?

atymic:
ping @lucasmichot

atymic:
ping @lucasmichot

lucasmichot:
> This allows v4 or v5, meaning if the user has v4, which doesn't have the grant type won't the provider break?
> I guess to merge this we'd need to bump the min to v5 on manager?

Sorry for the late reply. Yes indeed, we need to bump in the Manager package at some point

atymic:
v4 is two years old, and we'd only lose support for laravel 5.7/5.8
Happy to bump to v5 min then we can merge this?

https://github.com/laravel/socialite/releases/tag/v5.0.0

lucasmichot:
> v4 is two years old, and we'd only lose support for laravel 5.7/5.8 Happy to bump to v5 min then we can merge this?
> 
> https://github.com/laravel/socialite/releases/tag/v5.0.0

Totally fine with losing support for these versions 👍

atymic:
Tagged as v4.2, i guess you'll have to enforce a min version for every package here?
https://github.com/SocialiteProviders/Manager/releases/tag/v4.2.0

lucasmichot:
Thanks @atymic , see https://github.com/SocialiteProviders/Providers/pull/899

PR #884, opened 2022-08-06
Normalize the composer.json file for each package

lucasmichot:
This PR normalizes the `composer.json` file for each package


atymic (commented):
Could we get a github action to check this? So we dont have to do massive changes?

lucasmichot:
See https://github.com/SocialiteProviders/Providers/pull/896 @atymic

atymic:
Used the actions, thanks!

PR #887, opened 2022-08-10
Update PayPal to use latest endpoints / user info

jstowey:
PayPal have updated their documentation and the previous endpoints are no longer documented anywhere.

They've also made some slight changes to the data being returned: https://developer.paypal.com/docs/api/identity/v1/

### DIFF
--- a/src/PayPal/Provider.php
+++ b/src/PayPal/Provider.php
@@ -68,7 +68,7 @@ class Provider extends AbstractProvider
             'id'       => str_replace('https://www.paypal.com/webapps/auth/identity/user/', null, $user['user_id']),
             'nickname' => null, 'name' => $user['name'],
             'email'    => !empty($user['emails']) ? $user['emails'][0]['value'] : null,
-            'avatar' => null,
+            'avatar'   => null,
         ]);
     }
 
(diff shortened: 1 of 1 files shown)


src/PayPal/Provider.php
```diff
@@ -67,30 +67,24 @@ protected function mapUserToObject(array $user)
         return (new User())->setRaw($user)->map([
             'id'       => str_replace('https://www.paypal.com/webapps/auth/identity/user/', null, $user['user_id']),
             'nickname' => null, 'name' => $user['name'],
-            'email'    => $user['email'], 'avatar' => null,
+            'email'    => !empty($user['emails']) ? $user['emails'][0]['value'] : null,
```
atymic, line 74: Should we not use the primary email?  code below not tested

```suggestion
            'email'    => collect($user['emails'] ?? [])->firstWhere('primary', true),
```

atymic (changes requested):

src/PayPal/Provider.php
jstowey: Yea fair point, changed in latest commit

atymic (approved):

PR #891, opened 2022-08-17
PayPal - Catch missing user data error

jstowey:
It's possible "name" isn't returned from the API, so should default to an empty string

### DIFF
--- a/src/PayPal/Provider.php
+++ b/src/PayPal/Provider.php
@@ -67,7 +67,7 @@ class Provider extends AbstractProvider
     {
         return (new User())->setRaw($user)->map([
             'id'       => str_replace('https://www.paypal.com/webapps/auth/identity/user/', null, $user['user_id']),
-            'nickname' => null, 'name' => $user['name'],
+            'nickname' => null, 'name' => $user['name'] ?? '',
             'email'    => collect($user['emails'] ?? [])->firstWhere('primary')['value'] ?? null,
             'avatar'   => null,
         ]);


src/PayPal/Provider.php
```diff
@@ -67,7 +67,7 @@ protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
             'id'       => str_replace('https://www.paypal.com/webapps/auth/identity/user/', null, $user['user_id']),
-            'nickname' => null, 'name' => $user['name'],
+            'nickname' => null, 'name' => $user['name'] ?? '',
```
atymic, line 70: can we split nickname and name on their own lines?
also, I think null as the default makes more sense.

atymic (changes requested):

src/PayPal/Provider.php
jstowey: Sure thing

atymic (approved):

src/PayPal/Provider.php
```diff
@@ -67,7 +67,8 @@ protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
             'id'       => str_replace('https://www.paypal.com/webapps/auth/identity/user/', null, $user['user_id']),
```
atymic, line 69: Know this would be a breaking change, but why is the ID a url, and not just the user's id?

jstowey: Good question, wish I knew, but that's just how it's returned from the API: https://developer.paypal.com/docs/api/identity/v1/#userinfo_get

In my implementation I actually return the "payer_id" as the ID (which requires the "https://uri.paypal.com/services/paypalattributes" scope) and is much more useful, at least for my purposes, but I didn't want to make a breaking change.

PR #892, opened 2022-08-23
Update KakaoProvider.php to fIx getAccessToken return value

coldplay126:
Hi

i am a user appreciate your effort to keep this repository.

i think i found a code to need fix it.

"getAccessToken" function need to return string($code) so it using "parseAccessToken" function. but "parseAccessToken" function get need array but string before parsing json.

it might be a mistake because above the return code, there are already declare array to parse
($this->credentialsResponseBody).

could you please fix this issue?

sorry for my poor english.

sincerely

### DIFF
--- a/src/Kakao/KakaoProvider.php
+++ b/src/Kakao/KakaoProvider.php
@@ -48,7 +48,7 @@ class KakaoProvider extends AbstractProvider
 
         $this->credentialsResponseBody = json_decode((string) $response->getBody(), true);
 
-        return $this->parseAccessToken($response->getBody());
+        return $this->parseAccessToken($this->credentialsResponseBody);
     }
 
     /**


atymic:
Looks correct to me, have you tested the fix?

coldplay126:
@atymic 
yes i tested the code.

as you can see, ```parseAccessToken()``` method get an array params.

```
 /**
     * Get the access token from the token response body.
     *
     * @param  array  $body
     * @return string
     */
    protected function parseAccessToken($body)
    {
        return Arr::get($body, 'access_token');
    }
```

but ```$response->getBody()``` returns json format.

so without json_decode, ```parseAccessToken()``` can not take a ```$response->getBody()```. 

and i using  ```Socialite::driver('kakao')->user()```,  there are also json_decode the response.
```
// socialite\src\Two\AbstractProvider.php
 
// ...

public function user()
    {
        if ($this->user) {
            return $this->user;
        }

        if ($this->hasInvalidState()) {
            throw new InvalidStateException;
        }

        $response = $this->getAccessTokenResponse($this->getCode());

        $this->user = $this->mapUserToObject($this->getUserByToken(
            $token = Arr::get($response, 'access_token')
        ));

        return $this->user->setToken($token)
                    ->setRefreshToken(Arr::get($response, 'refresh_token'))
                    ->setExpiresIn(Arr::get($response, 'expires_in'))
                    ->setApprovedScopes(explode($this->scopeSeparator, Arr::get($response, 'scope', '')));
    }

// ... 

/**
     * Get the access token response for the given code.
     *
     * @param  string  $code
     * @return array
     */
    public function getAccessTokenResponse($code)
    {
            RequestOptions::HEADERS => $this->getTokenHeaders($code),
            RequestOptions::FORM_PARAMS => $this->getTokenFields($code),
        ]);

        return json_decode($response->getBody(), true); // <-- json_decode here
    }
```

atymic (approved):

PR #897, opened 2022-08-30
Add a GA to Lint and commit all code files

lucasmichot:
Add a GA to Lint and commit all code files; this workflow can be dispatched manually.


atymic:
Hmm, still not sure about this. This would probably update almost all files in the repo, making git blame a bit hard to use. Is there an issue with style ci? 🤔

PR #901, opened 2022-09-03
Saml2: Always provide the consumer URL in the AuthNRequest

### LINKED ISSUE #893: SAML 2.0 AuthN Request Missing ACS
Hello All,

Great library and I appreciate that it is kept up to date!

I have found what I think is a "bug" - and I say "bug" because while it doesn't break anything it also makes the library unusable in certain circumstances.

## What:
A SAML 2.0 AuthNRequest can - and is sometimes required depending on the configuration of the Identity Provider - to include the Assertion Consumer Service URL in the body of the request. However, this is not included, even when adding the `sp_acs` configuration in the `config/services.php` file per the instructions. Inspection of the request shows it is missing

## When:
All the time. The function used to generate the actual request simply does not include it.

## How:
Generate a AuthNRequest and the ACS is missing, regardless if a config entry is made

## Why:
This is not an issue if the Identity Provider does not request it or one has control over it to configure it as needed. However, when integrating with 3rd party based Identify Providers - think corporate SSO systems - it is possible they require it and there may be not way to turn it off. When this happens this library becomes unusable because of this.

## Possible fix
If I adding the following to `SocialiteProviders\Saml2\Provider.php` at line 190
```php
        if($this->getConfig('sp_acs')) {
            $authnRequest->setAssertionConsumerServiceURL($this->getAssertionConsumerServiceRoute());
        }
```
It will add the ACS to the AuthNReqeuest

## Bonus
I have also noticed that the `sp_acs` setting is automatically trimmed as well, cutting off any domain, always making the URL relative.  Again, this will work in some circumstances but possibly not in others. I would suggest allowing the use of the `sp_acs` URl as it is given - meaning if provided relative, keep it relative, if it is provided absolute, keep it absolute

If I changed the following at `SocialiteProviders\Saml2\Provider.php`  line 397
```php
return Str::of($this->getConfig('sp_acs', 'auth/callback'));
```
It will give both the relative and absoluste URL based on what I provided as a config

27pchrisl:
Fixes #893 - this adds the assertion consumer URL in the authentication request.

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -185,7 +185,8 @@ class Provider extends AbstractProvider implements SocialiteProvider
             ->setIssueInstant(new DateTime())
             ->setDestination($identityProviderConsumerService->getLocation())
             ->setNameIDPolicy((new NameIDPolicy())->setFormat(SamlConstants::NAME_ID_FORMAT_PERSISTENT))
-            ->setIssuer(new Issuer($this->getServiceProviderEntityDescriptor()->getEntityID()));
+            ->setIssuer(new Issuer($this->getServiceProviderEntityDescriptor()->getEntityID()))
+            ->setAssertionConsumerServiceURL($this->getServiceProviderAssertionConsumerUrl());
 
         if ($this->usesState()) {
             $this->request->session()->put('state', $state = $this->getState());


atymic:
@lucasmichot is the bot working correctly?

27pchrisl:
@atymic I am submitting these from an organization account. I'd prefer to keep doing that, but I can switch to a personal account if the new workflow requires it.

lucasmichot:
> @lucasmichot is the bot working correctly?

Yes it does.

Jason-Morcos:
Hi @27pchrisl. For some reason, this change made in the 4.4.0 release breaks my single sign on. It appears this is purely related to `SpSsoDescriptor-> getFirstAssertionConsumerService` being called without `SpSsoDescriptor->addAssertionConsumerService` being called previously. I believe this makes this PR a regression/breakign change to the saml2 provider

Screen Shot 2022-09-21 at 13 54 08

PR #908, opened 2022-09-09
Add config option for default algorithm JWK

esseremmerik:
The current AzureB2C provider does not contain an option to provide a default algorithm when none is provided by the provider. With this commit an additional config key is used to setup the default algorithm. It hooks into existing functionality of the JWK library which can handle a default setting.


atymic (approved):

atymic:
Could you update the docs (readme) as well please?

esseremmerik:
@atymic No problem, hope this update in the readme suits the requirements? If not, let me know.

atymic (approved):

Issue #909, opened 2022-09-13
Etsy sunsetting v2 for v3 OAuth2


cdburgess:
Are there any plans to upgrade the Etsy OpenApi to v3 since v2 is being sunsetted in Q4 2022?  I am happy to figure this out but I would rather not duplicate any work that may be coming.

atymic:
As far as I know, no. Feel free to PR.

cdburgess:
Pull request created. Please let me know if you have any questions. This has been tested on my local environment with my Etsy credentials and is working as expected.

PR #910, opened 2022-09-13
Etsy Open API v3 Update

cdburgess:
This is in response to issue #909 

The new OpenApi V3 now requires all calls to use OAuth2. The old v2 API is scheduled to be sunset in Q4 2022.

- This update removes the unneeded Server.php used for Oauth1 calls
- The README.md has been updated to provide a few more details about the calls and what is required
- Scopes are updated to a space-separated list as required by the API
- The `x-api-key` is included in the header as required by the API


src/Etsy/Provider.php
```diff
@@ -2,21 +2,82 @@
...
+            'grant_type' => 'authorization_code',
+        ]);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function formatScopes(array $scopes, $scopeSeparator)
+    {
+        return implode(' ', $scopes);
```
atymic, line 81: Set the scope separator to space instead, see other providers

src/Etsy/README.md
```diff
@@ -33,12 +43,25 @@ protected $listen = [
...
 ```
 
-### Usage
+### Usage `web/routes.php`
 
-You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed).
+However, there are a few things to note with the OpenApi v3 on Etsy. They require the use of PKCE. They also require that 
+`x-api-key` with your API Key be included in thea header for each request. The header code is taken care of in this provider.
+But you will need to call `enablePKCE` on your requests as well as add any scopes you want to access. The `email_r` is enabled
```
atymic, line 51: No need to do this, just set this on the provider then you can skip the enable method

```php
protected $usesPKCE = true;
```

src/Etsy/composer.json
```diff
@@ -1,18 +1,18 @@
 {
     "name": "socialiteproviders/etsy",
-    "description": "Etsy OAuth1 Provider for Laravel Socialite",
+    "description": "EtsyV3 OAuth2 Provider for Laravel Socialite",
```
atymic, line 3: Bump the manager please

"socialiteproviders/manager": "^4.2"

src/Etsy/Provider.php
```diff
@@ -2,21 +2,82 @@
...
+            'avatar'   => $user['image_url_75x75'],
         ]);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getTokenFields($code)
+    {
+        return array_merge(parent::getTokenFields($code), [
```
atymic, line 71: this is not needed if you bump the manager version to 4.2 see other comment

atymic (changes requested):

atymic:
Thanks for the PR, see comments :)

cdburgess:
Thanks for the suggestions! I have made all of those changes and retested them. Everything works as expected.

composer.json
```diff
@@ -11,7 +11,7 @@
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
         "league/oauth1-client": "^1.9",
-        "socialiteproviders/manager": "~4.0"
+        "socialiteproviders/manager": "^4.2"
```
atymic, line 14: please bump this for the provider not for the root composer json :)

cdburgess: Oops! Good catch. Sorry for the error.

atymic: Looks like its still in the root json

src/Etsy/EtsyExtendSocialite.php
```diff
@@ -9,10 +9,10 @@ class EtsyExtendSocialite
     /**
      * Register the provider.
      *
-     * @param \SocialiteProviders\Manager\SocialiteWasCalled $socialiteWasCalled
+     * @param SocialiteWasCalled $socialiteWasCalled
```
lucasmichot, line 12:
```suggestion
     * @param \SocialiteProviders\Manager\SocialiteWasCalled $socialiteWasCalled
```

src/Etsy/Provider.php
```diff
@@ -2,21 +2,74 @@
...
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
 
 class Provider extends AbstractProvider
 {
-    public const IDENTIFIER = 'ETSY';
+    /**
+     * Unique Provider Identifier.
+     */
+    const IDENTIFIER = 'ETSY';
```
lucasmichot, line 13:
```suggestion
    public const IDENTIFIER = 'ETSY';
```

src/Etsy/Provider.php
```diff
@@ -2,21 +2,74 @@
...
+    {
+        $tokenData = explode('.', $token);
+        $response = $this->getHttpClient()->get('https://openapi.etsy.com/v3/application/users/'.$tokenData[0], [
+            'headers' => [
+                'Authorization' => 'Bearer '.$token,
+                'x-api-key'     => $this->clientId,
+            ],
+        ]);
+
+        return json_decode($response->getBody(), true);
```
lucasmichot, line 59:
```suggestion
        return json_decode((string) $response->getBody(), true);
```

src/Etsy/README.md
```diff
@@ -33,12 +43,21 @@ protected $listen = [
...
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed).
 
-You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
+Note: The `email_r` is enabled by default so you can access user information in the callback. 
 
 ```php
-return Socialite::driver('etsy')->redirect();
+// the redirect
+return Socialite::driver('etsy')
+        ->scopes[['include','scopes','here']]
```
lucasmichot, line 55:
```suggestion
        ->scopes[['include', 'scopes', 'here']]  
```

lucasmichot (commented):
Thanks for the PR @cdburgess - Can you check my comments ?

src/Etsy/Provider.php
```diff
@@ -2,21 +2,74 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getUserByToken($token)
+    {
+        $tokenData = explode('.', $token);
+        $response = $this->getHttpClient()->get('https://openapi.etsy.com/v3/application/users/'.$tokenData[0], [
+            'headers' => [
```
lucasmichot, line 53: Can you use instead `\GuzzleHttp\RequestOptions::HEADERS` constant here instead ?

cdburgess: Updated!

composer.json
```diff
@@ -11,7 +11,7 @@
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
         "league/oauth1-client": "^1.9",
-        "socialiteproviders/manager": "~4.0"
+        "socialiteproviders/manager": "~4.2"
```
lucasmichot, line 14:
```suggestion
        "socialiteproviders/manager": "~4.0"
```

cdburgess: As an FYI, @atymic asked me to update this to `^4.2`... see his comments above. But I have downgraded as per your request.

composer.json
cdburgess: I believe this has been resolved now?

cdburgess:
Thanks for all the great input. I will check this again in the morning for any change requests.

lucasmichot (approved):
GTM @atymic

atymic (approved):

atymic:
Thanks @lucasmichot

atymic:
https://github.com/SocialiteProviders/Etsy/releases/tag/5.0.0

PR #918, opened 2022-09-23
Added Etsy Open API v3 Provider

rboydev:
Etsy is going to deprecate Open API v2. There is a new version of the API available which is covered by this provider. Version 2 is going to be available for a limited period of time.


atymic:
Isn't this done in https://github.com/SocialiteProviders/Providers/pull/910

rboydev:
Yes, it was done.
You can scratch this PR.

PR #919, opened 2022-09-26
Fix pinterest provider

AstroCorp:


atymic:
Can you clarify what this fixes?

AstroCorp:
This:
`$user = Socialite::driver('pinterest')->user();`

Show this error:
`Client error: POST https://api.pinterest.com/v5/oauth/token resulted in a 400 Bad Request response: {"code":1,"message":"Invalid request body"}`

I asked pinterest what was going on and this is their response:
Screenshot_20220926_114821

jsainton-godaddy:
FWIW, I think a lighter touch approach here would be to simply add the following code to the Provider:
```
    protected function getTokenHeaders($code)
    {
        return [
            'Authorization' => 'Basic '.base64_encode($this->clientId.':'.$this->clientSecret),
            'Content-Type' => 'application/x-www-form-urlencoded',
        ];
    }
```
Confirmed locally that this approach works fine, without removing any other values.

atymic:
@jsainton-godaddy could you  PR that please?

AstroCorp:
@atymic ready.

atymic (approved):

PR #920, opened 2022-10-03
Migrate from Patreon API v1 to v2.

### LINKED ISSUE #890: Patreon Provider return bad request when using v2 API
Hello, 

I'm using the Patreon Provider with v1 API keys and everything works fine. However, when I try with v2 API keys I get the following error : 

 ```
 Client error: `POST https://api.patreon.com/oauth2/token` resulted in a `400 Bad Request` response: {"error": "invalid_request", "error_description": "Missing code parameter."}
``` 

I don't see any documentation regarding v2 API on the provider, so I'm wondering if it's even supported?

sir-octopus:
Migrates from Patreon API v1 to v2. Fixes #890

### DIFF
--- a/src/Patreon/Provider.php
+++ b/src/Patreon/Provider.php
@@ -9,12 +9,16 @@ use SocialiteProviders\Manager\OAuth2\User;
 
 class Provider extends AbstractProvider
 {
+    /**
+     * Unique Provider Identifier.
+     */
     public const IDENTIFIER = 'PATREON';
 
     /**
      * {@inheritdoc}
      */
-    protected $scopes = ['users', 'pledges-to-me', 'my-campaign'];
+    // protected $scopes = ['users', 'pledges-to-me', 'my-campaign'];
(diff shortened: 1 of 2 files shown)


lucasmichot:
Thanks for this PR @sir-octopus , see [`fef5e80` (#920)](https://github.com/SocialiteProviders/Providers/pull/920/commits/fef5e80099cfb30cfc22695d51348c4b0676cbbf) for the code adjustments.

atymic:
I guess this needs a new major version if they api keys are different?

Jnewbon:
Testing might need to be done but according to the [patreon docs](https://docs.patreon.com/#apiv2-oauth)
> You will be able to request tokens with any set of APIv2 scopes from your existing APIv1 client. If you choose to create an APIv2-specific client in the developer portal, that client will only be able to request V2 scopes.

So V1 Client id and key will work with the V2 API fine

atymic (approved):

atymic:
I'm going to release it as major version bump just in case :)

PR #951, opened 2022-12-07
fix: php 8.1 compat twitter

recca0120:


atymic (commented):
Can you explain your change please?

ali-alharthi:
This should be reverted. 
`[Indirect modification of overloaded property SocialiteProviders\Manager\OAuth1\User::$urls has no effect]`

`PHP 8.2.6`

PR #953, opened 2022-12-10
Update Reddit provider with missing 'state' param

jeremykenedy:
Update Reddit provider with missing 'state' parameter.

It will fail without it with very little context.
* https://www.reddit.com/r/redditdev/comments/4hhtth/oauth2_that_option_is_not_valid/

state is random string.
* https://github.com/reddit-archive/reddit/wiki/OAuth2#authorization

### DIFF
--- a/src/Reddit/Provider.php
+++ b/src/Reddit/Provider.php
@@ -2,12 +2,14 @@
 
 namespace SocialiteProviders\Reddit;
 
-use GuzzleHttp\RequestOptions;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
 
 class Provider extends AbstractProvider
 {
+    /**
+     * Unique Provider Identifier.
+     */
     public const IDENTIFIER = 'REDDIT';
 
     /**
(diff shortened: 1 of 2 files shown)


src/Reddit/Provider.php
```diff
@@ -2,12 +2,14 @@
...
 namespace SocialiteProviders\Reddit;
 
-use GuzzleHttp\RequestOptions;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
 
 class Provider extends AbstractProvider
 {
+    /**
+     * Unique Provider Identifier.
```
atymic, line 11: No need for this

src/Reddit/Provider.php
```diff
@@ -26,6 +28,33 @@ protected function getAuthUrl($state)
...
+            '',
+            '&',
+            $this->encodingType
+        );
+    }
+
+    /**
+     * @param string $arrayKey
+     */
+    protected function getFromConfig($arrayKey)
```
atymic, line 53: Why was this added? You should be able to get config keys already

src/Reddit/Provider.php
```diff
@@ -42,14 +71,14 @@ protected function getUserByToken($token)
         $response = $this->getHttpClient()->get(
             'https://oauth.reddit.com/api/v1/me',
             [
-                RequestOptions::HEADERS => [
+                'headers' => [
```
atymic, line 74: Can you leave this as the constant?

src/Reddit/Provider.php
```diff
@@ -42,14 +71,14 @@ protected function getUserByToken($token)
...
                     'Authorization' => 'Bearer '.$token,
                     'User-Agent'    => $this->getUserAgent(),
                 ],
             ]
         );
 
-        return json_decode((string) $response->getBody(), true);
+        return json_decode($response->getBody()->getContents(), true);
```
atymic, line 81: Again, this and below is correct already?

src/Reddit/README.md
```diff
@@ -14,6 +14,8 @@ Please see the [Base Installation Guide](https://socialiteproviders.com/usage/),
 'reddit' => [    
   'client_id' => env('REDDIT_CLIENT_ID'),  
   'client_secret' => env('REDDIT_CLIENT_SECRET'),  
+  'response_type' => env('REDDIT_RESPONSE_TYPE', 'code'),
+  'state' => env('REDDIT_STATE', 'r@nd0m5tr1n6'),
```
atymic, line 18: This is not how state works, it should be randomly generated each request. Have you told the provider to be stateless? Otherwise it should be generated automatically.

atymic (changes requested):
See comments

Screen Shot 2022-12-12 at 08 59 13

src/Reddit/README.md
jeremykenedy: Yeah, I'm doing stateless.

PR #960, opened 2022-12-24
Update Instagram provider to match Instagram api and work again.

jeremykenedy:
Update work out the box the API calls in `getUserByToken` were calling deprecated methods.

### DIFF
--- a/src/Instagram/Provider.php
+++ b/src/Instagram/Provider.php
@@ -2,6 +2,7 @@
 
 namespace SocialiteProviders\Instagram;
 
+use GuzzleHttp\RequestOptions;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
 use SocialiteProviders\Manager\OAuth2\User;
 
@@ -22,7 +23,7 @@ class Provider extends AbstractProvider
      *
      * @var array
      */
-    protected $fields = ['account_type', 'id', 'username', 'media_count', 'profile_picture'];
+    protected $fields = ['account_type', 'id', 'username', 'media_count'];
 
     /**
      * {@inheritdoc}
@@ -60,7 +61,7 @@ class Provider extends AbstractProvider
             $meUrl .= '&appsecret_proof='.$appSecretProof;
         }
         $response = $this->getHttpClient()->get($meUrl, [
-            'headers' => [
+            RequestOptions::HEADERS => [
                 'Accept' => 'application/json',
             ],
         ]);
@@ -77,7 +78,6 @@ class Provider extends AbstractProvider
             'id'            => $user['id'],
             'name'          => $user['username'],
             'account_type'  => $user['account_type'],
-            'avatar'        => $user['profile_picture'] ?? null,
             'media_count'   => $user['media_count'] ?? null,
         ]);
     }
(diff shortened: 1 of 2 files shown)


src/Instagram/Provider.php
```diff
@@ -70,19 +74,21 @@ protected function getUserByToken($token)
...
+            'avatar'        => $user['profile_picture'] ?? null,
+            'media_count'   => $user['media_count'] ?? null,
         ]);
     }
 
     public function getAccessToken($code)
     {
         $response = $this->getHttpClient()->post($this->getTokenUrl(), [
-            RequestOptions::FORM_PARAMS => $this->getTokenFields($code),
+            'form_params' => $this->getTokenFields($code),
```
atymic, line 88: Why was this changed?
https://docs.guzzlephp.org/en/stable/request-options.html#form-params

src/Instagram/Provider.php
```diff
@@ -70,19 +74,21 @@ protected function getUserByToken($token)
...
         ]);
 
-        $this->credentialsResponseBody = json_decode((string) $response->getBody(), true);
```
atymic, line 85: Same here, this is correct.

src/Instagram/Provider.php
```diff
@@ -44,24 +53,19 @@ protected function getTokenUrl()
...
-        $response = $this->getHttpClient()->get(
-            'https://api.instagram.com/v1/users/self',
-            [
-                RequestOptions::QUERY   => $query,
-                RequestOptions::HEADERS => [
-                    'Accept' => 'application/json',
-                ],
-            ]
-        );
+        $meUrl = 'https://graph.instagram.com/me?access_token='.$token.'&fields='.implode(',', $this->fields);
```
atymic, line 56: Can you link to the docs? this look a major change in API

src/Instagram/Provider.php
```diff
@@ -2,23 +2,32 @@
...
      * {@inheritdoc}
      */
     protected $scopeSeparator = ' ';
 
+    /**
+     * The user fields being requested.
+     *
+     * @var array
+     */
+    protected $fields = ['account_type', 'id', 'username', 'media_count', 'profile_picture'];
```
atymic, line 25: Do these need to be customisable?

atymic (changes requested):

src/Instagram/Provider.php
jeremykenedy: https://developers.facebook.com/docs/instagram-basic-display-api/reference/me

https://developers.facebook.com/docs/instagram-basic-display-api/reference/user

https://developers.facebook.com/docs/instagram-basic-display-api/reference/oauth-authorize

src/Instagram/Provider.php
jeremykenedy: This was an oversite and has been reverted back.

src/Instagram/Provider.php
jeremykenedy: This was an oversite and has been reverted back.

src/Instagram/Provider.php
jeremykenedy: No, those are the only fields available. https://developers.facebook.com/docs/instagram-basic-display-api/reference/user#fields

src/Instagram/Provider.php
jeremykenedy: Meta is trying to discourage people from using login with IG and directing them to use FB. So the data is limited. the previous endpoint of `/users/self` no longer exists and will 404. I believe they replaced it with `/me`.

atymic (approved):

PR #962, opened 2023-01-04
[Apple] Adding method to return SocialiteUser by JWT token

guillaumeboussion:
Because `getUserByToken()` is a protected function, you cannot check an identity token directly from the Socialite driver. This PR intend to allow the verification of a JWT by a public function, returning a SocialiteUser, or throwing an InvalidStateException when the given JWT is incorrect

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -122,6 +122,7 @@ class Provider extends AbstractProvider
      * side by Apple.
      *
      * @param string $token
+     * 
      * @throws InvalidStateException when token can't be parsed
      *
      * @return User $user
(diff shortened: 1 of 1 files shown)


atymic (changes requested):
Hmm, throwing invalidstateexception feels wrong. Maybe change the verify method exception, and catch and re-throw as invalid state in the actual signin?

guillaumeboussion:
The thing is, the provider's `verify` method is already throwing this exception when JWK constraints are not satisfied. Which kind of exception has to be thrown in this case ?

atymic (approved):

PR #965, opened 2023-01-13
[Dataporten] add openid scope and add logout functionality

0hra:
Relevant docs for "openid" scope (see "Authorization Request"):
[Obtaining tokens](https://docs.feide.no/service_providers/openid_connect/feide_obtaining_tokens.html)
If scope not present, you will not obtain "id_token" which is needed for logout functionality.

Relevant docs for logout:
[Redirection after logout](https://docs.feide.no/service_providers/manage/openid_connect/redir_etter_logout.html)

### DIFF
--- a/src/Dataporten/Provider.php
+++ b/src/Dataporten/Provider.php
@@ -12,7 +12,7 @@ class Provider extends AbstractProvider
 
     /**
      * {@inheritdoc}
-    */
+     */
     protected $scopes = ['openid'];
 
     /**
@@ -75,5 +75,4 @@ class Provider extends AbstractProvider
 
         return "$endpointUri?$params";
     }
-
 }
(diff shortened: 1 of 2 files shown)


src/Dataporten/Provider.php
```diff
@@ -57,4 +62,17 @@ protected function getTokenFields($code)
             'grant_type' => 'authorization_code',
         ]);
     }
+
+    /**
+     *  Get the endsession endpoint URL.
+     */
+    public function getLogoutUrl($idToken, $endpointUri, $redirectUri)
+    {
+        $params = http_build_query(array_filter([
```
atymic, line 71: This method doesn't really do anything. Could we put some of these field as config, and the have the provider build it automatically?

atymic: Could we also get strict types please

atymic (changes requested):

atymic:
Closing due to inactivity. Feel free to ping me if once you make the changes.

PR #971, opened 2023-02-02
[Microsoft] Added ability to retrieve tenant information when using multi-tenant auth

allw1994:
Added ability to retrieve tenant information when using multi-tenant auth to allow for tenant filtering for any apps that may wish to form allow/blocklists or who may want to only allow access to "internal" tenants.

### DIFF
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -79,7 +79,7 @@ class Provider extends AbstractProvider
 
         $formatted_response = json_decode((string) $responseUser->getBody(), true);
 
-        if (($this->getConfig('tenant', 'common') === 'common') && ($this->getConfig('include_tenant_info', 'false') === 'true')) {
+        if (($this->getConfig('tenant', 'common') === 'common') && ($this->getConfig('include_tenant_info', false) === true)) {
             $responseTenant = $this->getHttpClient()->get(
                 'https://graph.microsoft.com/v1.0/organization',
                 [
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -72,7 +72,7 @@ class Provider extends AbstractProvider
                     'Authorization' => 'Bearer '.$token,
                 ],
                 RequestOptions::QUERY => [
-                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_USER, ($this->config['fields'] ?: []))),
+                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_USER, $this->getConfig('fields'))),
                 ],
             ]
         );
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -88,7 +88,7 @@ class Provider extends AbstractProvider
                         'Authorization' => 'Bearer '.$token,
                     ],
                     RequestOptions::QUERY => [
-                        '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_TENANT, ($this->config['tenant_fields'] ?: []))),
+                        '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_TENANT, $this->getConfig('tenant_fields'))),
                     ],
                 ]
             );
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -79,7 +79,7 @@ class Provider extends AbstractProvider
 
         $formatted_response = json_decode((string) $response_user->getBody(), true);
 
-        if($this->getConfig('tenant','common') === 'common'){
+        if($this->getConfig('tenant','common') === 'common') {
             $response_tenant = $this->getHttpClient()->get(
                 'https://graph.microsoft.com/v1.0/organization',
                 [
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -79,7 +79,7 @@ class Provider extends AbstractProvider
 
         $formattedResponse = json_decode((string) $responseUser->getBody(), true);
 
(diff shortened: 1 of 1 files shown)


src/Microsoft/Provider.php
```diff
@@ -58,20 +64,39 @@ protected function getTokenUrl()
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get(
+        $response_user = $this->getHttpClient()->get(
```
atymic, line 67:
```suggestion
        $responseUser = $this->getHttpClient()->get(
```

Camelcase please

atymic: (and in other spots)

src/Microsoft/Provider.php
```diff
@@ -58,20 +64,39 @@ protected function getTokenUrl()
...
-                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS, ($this->config['fields'] ?: []))),
+                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_USER, ($this->config['user_fields'] ?: []))),
                 ],
             ]
         );
 
-        return json_decode((string) $response->getBody(), true);
+        $formatted_response = json_decode((string) $response_user->getBody(), true);
+
+        if ($this->getConfig('tenant', 'common') === 'common') {
```
atymic, line 82: should we not make this opt in, rather than opt out?

atymic (changes requested):

src/Microsoft/Provider.php
allw1994: Have included in further changes.

allw1994:
Does anything else still need to be tweaked?

src/Microsoft/Provider.php
```diff
@@ -58,20 +64,39 @@ protected function getTokenUrl()
...
                 ],
                 RequestOptions::QUERY => [
-                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS, ($this->config['fields'] ?: []))),
+                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_USER, ($this->config['fields'] ?: []))),
                 ],
             ]
         );
 
-        return json_decode((string) $response->getBody(), true);
+        $formatted_response = json_decode((string) $responseUser->getBody(), true);
```
atymic, line 80: camelcase please

src/Microsoft/Provider.php
```diff
@@ -58,20 +64,39 @@ protected function getTokenUrl()
...
+
+        if (($this->getConfig('tenant', 'common') === 'common') && ($this->getConfig('include_tenant_info', false) === true)) {
```
atymic, line 82:
```suggestion
        if ($this->getConfig('tenant', 'common') === 'common' && $this->getConfig('include_tenant_info', false) {
```

src/Microsoft/Provider.php
```diff
@@ -58,20 +64,39 @@ protected function getTokenUrl()
...
+                        'Accept'        => 'application/json',
+                        'Authorization' => 'Bearer '.$token,
+                    ],
+                    RequestOptions::QUERY => [
+                        '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_TENANT, ($this->config['tenant_fields'] ?: []))),
+                    ],
+                ]
+            );
+
+            $formatted_response['tenant'] = json_decode((string) $responseTenant->getBody(), true)['value'][0];
```
atymic, line 96: as above. Do we need a null coalesce here to make sure this does't error if the request doesn't have results?
```suggestion
            $formatted_response['tenant'] = json_decode((string) $responseTenant->getBody(), true)['value'][0] 
            ?? null;
```

atymic (changes requested):

src/Microsoft/Provider.php
```diff
@@ -58,20 +64,39 @@ protected function getTokenUrl()
...
+        if ($this->getConfig('tenant', 'common') === 'common' && $this->getConfig('include_tenant_info', false)) {
+            $responseTenant = $this->getHttpClient()->get(
+                'https://graph.microsoft.com/v1.0/organization',
+                [
+                    RequestOptions::HEADERS => [
+                        'Accept'        => 'application/json',
+                        'Authorization' => 'Bearer '.$token,
+                    ],
+                    RequestOptions::QUERY => [
+                        '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_TENANT, ($this->config['tenant_fields'] ?: []))),
```
atymic, line 91: Sorry one more 🙈 
```suggestion
                        '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_TENANT, $this->getConfig('tenant_fields'))),
```

src/Microsoft/Provider.php
```diff
@@ -58,20 +64,39 @@ protected function getTokenUrl()
...
+        $responseUser = $this->getHttpClient()->get(
             'https://graph.microsoft.com/v1.0/me',
             [
                 RequestOptions::HEADERS => [
                     'Accept'        => 'application/json',
                     'Authorization' => 'Bearer '.$token,
                 ],
                 RequestOptions::QUERY => [
-                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS, ($this->config['fields'] ?: []))),
+                    '$select' => implode(',', array_merge(self::DEFAULT_FIELDS_USER, ($this->config['fields'] ?: []))),
```
atymic, line 75: And may as well update it here as well

allw1994:
Done, I hope - want to make sure I write it into the docs and I have more to add to this file too so need to square this off first.

atymic (approved):

PR #972, opened 2023-02-06
Add USOS OAuth1.0a provider

m0stwan1ed:
Provider for [USOS](https://usos.edu.pl). This is a polish universal university management system, which also allows students and teachers to log in via OAuth 1.0a to a third-party applications

Developer docs: https://apps.usos.edu.pl/developers/api/authorization/

### DIFF
--- a/src/Usos/Provider.php
+++ b/src/Usos/Provider.php
@@ -22,4 +22,9 @@ class Provider extends AbstractProvider
             'avatar'   => $user['avatar'],
         ]);
     }
+
+    public static function additionalConfigKeys()
+    {
+        return ['profile_fields_selector', 'domain'];
+    }
 }
--- a/src/Usos/Provider.php
+++ b/src/Usos/Provider.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace SocialiteProviders\Usos;
+
+use SocialiteProviders\Manager\OAuth1\AbstractProvider;
+use SocialiteProviders\Manager\OAuth1\User;
+
+class Provider extends AbstractProvider
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    public const IDENTIFIER = 'USOS';
+
+    protected function mapUserToObject(array $user)
+    {
+        return (new User())->setRaw($user['extra'])->map([
+            'id'       => $user['id'],
+            'nickname' => $user['nickname'],
+            'name'     => $user['name'],
+            'email'    => $user['email'],
+            'avatar'   => $user['avatar'],
+        ]);
+    }
+}
--- a/src/Usos/README.md
+++ b/src/Usos/README.md
@@ -1,0 +1,65 @@
+# Usos
+
+```bash
+composer require socialiteproviders/usos
+```
+
+## Installation & Basic Usage
+
+Please see the [Base Installation Guide](https://socialiteproviders.com/usage/), then follow the provider specific instructions below.
+
+### Add configuration to `config/services.php`
+
+```php
+'usos' => [
+  'domain' => env('USOS_DOMAIN_URL'), // Because of every instance of USOS is self-hosted
+  'client_id' => env('USOS_CLIENT_ID'),
+  'client_secret' => env('USOS_CLIENT_SECRET'),
+  'redirect' => env('USOS_REDIRECT_URI')
+],
+```
+
+As a default, provider loads only necessary for Socialite data. If you need to get more data during authentication process, such as student status or phone number, you may add ```profile_fields_selector``` parameter:
+
+```php
+'usos' => [
+  // ...
+  'profile_fields_selector' => env('USOS_PROFILE_FIELDS_SELECTOR'),
+  // ...
+]
+```
+
+Some fields may be not available without ```scopes``` field specifying.
+
+### Add provider event listener
+
+Configure the package's listener to listen for `SocialiteWasCalled` events.
+
+Add the event to your `listen[]` array in `app/Providers/EventServiceProvider`. See the [Base Installation Guide](https://socialiteproviders.com/usage/) for detailed instructions.
+
+```php
+protected $listen = [
+    \SocialiteProviders\Manager\SocialiteWasCalled::class => [
+        // ... other providers
+        'SocialiteProviders\\Usos\\UsosExtendSocialite@handle',
+    ],
+];
+```
+
+### Usage
+
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
+
+```php
+return Socialite::driver('usos')->redirect();
+```
+
+You also can use USOS scopes via Socialite ```scopes``` method:
+
+```php
+return Socialite::driver('usos')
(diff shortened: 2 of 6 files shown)


src/Usos/composer.json
```diff
@@ -0,0 +1,34 @@
...
+            "email": "mostovyv@uek.krakow.pl"
+        }
+    ],
+    "support": {
+        "issues": "https://github.com/socialiteproviders/providers/issues",
+        "source": "https://github.com/socialiteproviders/providers",
+        "docs": "https://socialiteproviders.com/usos"
+    },
+    "require": {
+        "php": "^7.4 || ^8.0",
```
atymic, line 25:
```suggestion
        "php": "^8.0",
```

src/Usos/Server.php
```diff
@@ -0,0 +1,162 @@
...
+            $this->handleTemporaryCredentialsBadResponse($e);
+        }
+
+        throw new CredentialsException('Failed to get temporary credentials');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function userUid($data, TokenCredentials $tokenCredentials)
```
atymic, line 121: What are these functions returned for?

src/Usos/Provider.php
```diff
@@ -0,0 +1,25 @@
...
+
+use SocialiteProviders\Manager\OAuth1\AbstractProvider;
+use SocialiteProviders\Manager\OAuth1\User;
+
+class Provider extends AbstractProvider
+{
+    /**
+     * Unique Provider Identifier.
+     */
+    public const IDENTIFIER = 'USOS';
```
atymic, line 13: Does this need an oauth 2 provider?

src/Usos/README.md
```diff
@@ -0,0 +1,65 @@
...
+  'redirect' => env('USOS_REDIRECT_URI')
+],
+```
+
+As a default, provider loads only necessary for Socialite data. If you need to get more data during authentication process, such as student status or phone number, you may add ```profile_fields_selector``` parameter:
+
+```php
+'usos' => [
+  // ...
+  'profile_fields_selector' => env('USOS_PROFILE_FIELDS_SELECTOR'),
```
atymic, line 27: Don't see where this is used?

m0stwan1ed: According to the [USOS API documentation](https://apps.usos.edu.pl/developers/api/services/users/#user), only ID, first name and last name will be returned, if 'fields' selector wasn't provided.

According to the [Laravel Socialite Docs](https://laravel.com/docs/9.x/socialite#retrieving-user-details) Socialite driver must provide this set of data: id, name, email, and avatar. Email address and avatar (user's photo) won't be returned by API endpoint by default, so request must contain 'fields' parameter to get all necessary data (minimum: id|first_name|last_name|email|photo_urls). 

Because of the large list of possible fields, which could be returned, I've decided to allow application to re-define the list of extra parameters, which application need to get. For example: authentication only could be possible for students. In this situation, application can ask about a student status of user (student_status field from selector).

UPD: Sorry, I forgot to include this change to Server.php. I'll add it in the next commit

src/Usos/Provider.php
m0stwan1ed: I'm not quite understand the question.

USOS service only provides an OAuth1.0a authentication. Of course, [there is a 'oauth2' branch](https://apps.usos.edu.pl/developers/api/services/oauth2/), but it was named as '2' only because they want to add another (second) version of their implementation of OAuth1.0, not the OAuth2.0.

If you mean "Why do you add this lines?": I've added them, because another OAuth1.0 drivers (Such as Twitter) also has this const field. If I am wrong, please leave me a suggestion, how can I improve this.

src/Usos/Server.php
m0stwan1ed: Because of Server class extends BaseServer class, which based on an abstract Server class of League/OAuth1-client with abstract userUid method, the definition of this function is necessary.

According to the docs of league/oauth1-client, this function must return only unique authenticated user's id from list of user's data, which was returned from the user details URL. In this case, it takes user's unique ID from USOS data list, ([services/users/user](https://apps.usos.edu.pl/developers/api/services/users/#user)) and returns it.

atymic: All good, I have not reviewed many v1 providers 😅

src/Usos/Provider.php
atymic: I mean, is this class ever used? do we need it if the provider only supports v1?

src/Usos/README.md
atymic: Haha yea I didn't see it used in the code so was confused. Thanks!

src/Usos/Server.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private function getDomain()
+    {
+        return config('services.usos.domain');
+    }
+
+    private function getFieldsSelector()
+    {
+        return config('services.usos.profile_fields_selector', 'id|first_name|last_name|email|photo_urls');
```
atymic, line 53: Can we prefer the build in config helper please?

```php
    {
        return ['profile_fields_selector', 'domain'];
    }
```

```php
$this->getConfig('profile_fields_selector', 'id|first_name|last_name|email|photo_urls')
```

src/Usos/Server.php
```diff
@@ -0,0 +1,167 @@
...
+        $user = new User();
+        $user->id = $data['id'];
+        $user->nickname = $data['first_name'].' '.$data['last_name'];
+        $user->name = $data['first_name'].' '.$data['last_name'];
+        $user->avatar = $data['photo_urls']['50x50'] ?? null;
+        $user->email = $data['email'] ?? null;
+
+        $used = ['id', 'first_name', 'last_name', 'email'];
+
+        $user->extra = array_diff_key($data, array_flip($used));
```
atymic, line 70: I don't think we need this to filter the keys. Just use `setRaw` to set all the keys?

src/Usos/Provider.php
m0stwan1ed: According to the [Contribution Guide page of the project](https://socialiteproviders.com/contribute/#creating-a-provider), Every auth driver must have a handler/listener and Provider class. If it is a driver for OAuth1.0 service, it also must have a Server class.

Also, i've tried to remove that class, and I've got an error: `Call to undefined method App\Helpers\UsosSocialiteExtension\Provider::mapUserToObject()`.  I am not calling this method anywhere in the driver, so this method called automatically by Laravel Socialite during the handling of callback.

Everything says, that the extending of AbstractProvider class is necessary.

src/Usos/README.md
m0stwan1ed: I've fixed it in the next commit. It was automatically attached to this PR. If you let me, I'll mark this as a `resolved`

src/Usos/Server.php
m0stwan1ed: Thanks for the advice. I will add it with next push.

atymic (approved):

PR #977, opened 2023-02-16
[Apple] Use provider's client instance

axlon:
The way this provider makes the HTTP call to retrieve Apple's signing keys currently makes it very hard to test around this without actually making an HTTP call. Additionally, any configuration set on the provider's HTTP client will be ignored because it uses a fresh, unconfigured Guzzle instance.


axlon:
I just noticed this method is static, so it won't work

PR #979, opened 2023-02-20
[FusionAuth] Changing required properties and better handling API responses

vcampitelli:
Minor changes to the FusionAuth provider:

- :alien: Checking for either `name` or `given_name`/`family_name` attributes when building `User` object (the API may return both)
- :necktie: Making `tenant_id` optional (you can leave it blank if you only have one)
- :pencil2: Fixing typo with the `::class` constant in the README file (it was missing one `:`)

### DIFF
--- a/src/FusionAuth/Provider.php
+++ b/src/FusionAuth/Provider.php
@@ -89,10 +89,19 @@ class Provider extends AbstractProvider
      */
     protected function mapUserToObject(array $user)
     {
+        $name = null;
+        if (!empty($user['name'])) {
+            $name = $user['name'];
+        } elseif (!empty($user['given_name'])) {
+            $name = $user['given_name'];
+            if (!empty($user['family_name'])) {
+                $name .= " {$user['family_name']}";
+            }
+        }
         return (new User())->setRaw($user)->map([
             'id'       => $user['sub'],
             'nickname' => $user['preferred_username'] ?? null,
-            'name'     => $user['name'],
+            'name'     => $name,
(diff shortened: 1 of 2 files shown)


src/FusionAuth/Provider.php
```diff
@@ -89,10 +89,19 @@ protected function getUserByToken($token)
      */
     protected function mapUserToObject(array $user)
     {
+        $name = null;
+        if (!empty($user['name'])) {
```
atymic, line 93:
```suggestion
        $user['name'] ?? trim($user['given_name'] ?? '' . ' ' . $user['family_name'] ?? '')
```

atymic (approved):

atymic:
cc @vcampitelli

src/FusionAuth/Provider.php
```diff
@@ -89,10 +89,19 @@ protected function getUserByToken($token)
      */
     protected function mapUserToObject(array $user)
     {
+        $name = null;
+        $user['name'] ?? trim($user['given_name'] ?? '' . ' ' . $user['family_name'] ?? '')
```
atymic, line 93: Sorry i mean change the whole statement here to the oneliner i suggested

atymic (changes requested):

atymic:
@vcampitelli

vcampitelli:
Yes, sorry. That makes more sense :smile: Pushed again. Thanks.

atymic (approved):

PR #984, opened 2023-03-01
[Saml2] Allow the administrator to choose an entity descriptor to use

### LINKED ISSUE #961: SAML2 driver problem with multiple EntityDescriptors
**Problem**

I have metadata that includes multiple entitydescriptors for different authentication providers and the one I have to use is not the first one in that file. SAML2 driver only fetches the first descriptor and uses data from that so the login fails.

**Solution**

If user provides both metadata (url or a xml-file) AND entityid config values, then only fetch entitydescriptor for that entityid. If none are found or if entityid config value is not set then return the first one. 
It does not break current behaviour and supports situations where you have to choose the correct entitydescriptor.

27pchrisl:
Fixes #961

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -267,20 +267,33 @@ class Provider extends AbstractProvider implements SocialiteProvider
         return $builder->get();
     }
 
-    protected function getFirstEntityDescriptorFromXml(string $xml): EntityDescriptor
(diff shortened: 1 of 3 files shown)


atymic:
I assume there's no breaking changes relevant in the major version bump of lightsaml?

27pchrisl:
There are not, no. They replaced the symfony/event-dispatcher dependency with psr/event-dispatcher (I assume to make litesaml less symfony-specific).

PR #987, opened 2023-03-03
[Saml2] Configurable SAML2 Name ID Formats

samuel-fonseca:
While working on a project I came across an issue with the Name ID Format not being a match with the IDP's. I believe this is a valuable addition to the library to extend the configuration and add the ability to configure the `SamlConstants::NAME_ID_FORMAT_PERSISTENT` being used in the Provider class.

By default the value will continue to be the default `urn:oasis:names:tc:SAML:2.0:nameid-format:persistent` value. But it also allows the developer to set a new config value (`sp_name_id_format`) which would then change the name format.

I added a getter method, `Provider::getNameIDFormat()`, which validates the configured name id, in case it is not valid, it returns the `SamlConstants::NAME_ID_FORMAT_PERSISTENT` by default. Then updated the two references to use the getter instead of the constant directly.

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -171,6 +171,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
             'sp_org_display_name',
             'sp_org_url',
             'sp_default_binding_method',
+            'sp_name_id_format',
             'idp_binding_method',
             'attribute_map',
         ];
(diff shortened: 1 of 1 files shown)


27pchrisl:
Hi @samuel-fonseca, thanks for the PR! I agree this would be a useful addition.

The code as written silently ignores an invalid configuration, reverting to the default. This could be confusing for someone experimenting with that parameter - could you throw a `SocialiteProviders\Manager\Exception\MissingConfigException` with a helpful message if `SamlConstants::isNameIdFormatValid()` returns false?

samuel-fonseca:
@27pchrisl I have made those changes!

atymic (approved):

PR #994, opened 2023-03-16
get Tiktok user from the official user details endpoint

lebedevsergey:
The code in master branch gets user details through `open-api.tiktok.com/oauth/userinfo` endpoint which now seems outdated as it is not even mentioned in the current version of official Tiktok API [documentation](https://developers.tiktok.com/doc). Also with this endpoint I've faced issues like 'Undefined array key "avatar_larger"'. 
Furthermore using of this endpoint leads to code inconsistency because with the API access token it requires also to send `open_id` field in `getUserByToken` method.

At the same time there is an official [user details endpoint](https://developers.tiktok.com/doc/tiktok-api-v2-get-user-info) which requires only access token. 
The fix switches getUserByToken method to this endpoint.

### DIFF
--- a/src/TikTok/Provider.php
+++ b/src/TikTok/Provider.php
@@ -55,10 +55,7 @@ class Provider extends AbstractProvider
         $token = Arr::get($response, 'data.access_token');
 
         $this->user = $this->mapUserToObject(
-            $this->getUserByToken([
-                'access_token' => $token,
-                'open_id'      => Arr::get($response, 'data.open_id'),
-            ])
+            $this->getUserByToken($token)
         );
 
         return $this->user->setToken($token)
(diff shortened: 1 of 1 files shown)


atymic (approved):

IsraelOrtuno:
After updating I started getting this error:

```Undefined array key "open_id"```

Before using this new endpoint, the tiktok response was like:

```json
{
    "data": {
        "access_token": "act.c39724355c65afcd2d7026dcdbd76d40LJQ7fHsqvfUVmZl5nxWKEJ1FU1wc!6393",
        "captcha": "",
        "desc_url": "",
        "description": "",
        "error_code": 0,
        "expires_in": 86400,
        "log_id": "20220714041044010002007735002037040BAF34",
        "open_id": "cdbbfaea-34f6-4a45-ab45-b12f78b1813d",
        "refresh_expires_in": 31536000,
        "refresh_token": "rft.dc8946222f5c87a21f681dc7995e3hEuMiySL8qLspsKgovFb6D0NyMly!6418",
        "scope": "user.info.basic"
    },
    "message": "success"
}
```

But using the new endpoint, the user is nested under a `user` property:

```json
{
   "data":{
      "user":{
         "avatar_url":"https://p19-sign.tiktokcdn-us.com/tos-useast5-avt-0068-tx/b17f0e4b3a4f4a50993cf72cda8b88b8~c5_168x168.jpeg",
         "open_id":"723f24d7-e717-40f8-a2b6-cb8464cd23b4",
         "union_id":"c9c60f44-a68e-4f5d-84dd-ce22faeb0ba1",
         "display_name": "Tik Toker"
      }
   },
   "error":{
      "code":"ok",
      "message":"",
      "log_id":"20220829194722CBE87ED59D524E727021"
   }
}
```

So when calling `mapUserToObject`, it's accessing `data` but not `data.user` and it fails while extracting any parameter: https://github.com/SocialiteProviders/Providers/blob/master/src/TikTok/Provider.php#L108-L118 

Not sure if I am missing something and this was made in purpose or I need to so something else into order to fix it. Happy to open a PR if this is a bug.

PR #1005, opened 2023-04-17
fix(v2): migration to v2 of the streamlabs oauth api

AnkhHeart:


atymic:
Can we get a link to the docs please?

hamrak:
> Can we get a link to the docs please?

https://dev.streamlabs.com/docs

PR #1006, opened 2023-04-19
Add support for lcobucci/jwt 5.0

hamrak:


src/Apple/composer.json
```diff
@@ -40,7 +40,8 @@
         "ext-json": "*",
         "ext-openssl": "*",
         "firebase/php-jwt": "^6.2",
-        "lcobucci/jwt": "^4.1.5",
+        "lcobucci/clock": "^2.0 || ^3.0",
+        "lcobucci/jwt": "^4.1.5 || ^5.0.0",
```
atymic, line 44: Are there any breaking changes here?

hamrak: Here is list of all official BC-break changes of lcobucci/jwt for upgrate from v.4 to v.5: 

- [979: Builder: make it immutable](https://github.com/lcobucci/jwt/pull/979)
- [939: Remove empty Signer, empty Key, empty Signature, empty strings](https://github.com/lcobucci/jwt/pull/939)
- [937: Remove \Lcobucci\JWT\Signer\None](https://github.com/lcobucci/jwt/issues/937)
- [873: Use abstraction instead of concretion](https://github.com/lcobucci/jwt/pull/873)
- [872: Remove deprecated components](https://github.com/lcobucci/jwt/pull/872)
- [871: Require PHP 8.1+](https://github.com/lcobucci/jwt/pull/871)
- [969: 4.3.x merge up into 5.0.x](https://github.com/lcobucci/jwt/pull/969)

Minimum PHP for v.5 is 8.1, that's why I kept using the older version as well "^4.1.5 || ^5.0.0", but functionality in this package is compatible with both versions.

atymic (approved):

src/Apple/composer.json
atymic: Thanks!

PR #1007, opened 2023-04-21
Updated SuperOffice provider

SuperDJ:
Due to some upcoming updates in SuperOffice they require that the API endpoint URL is dynamic.
- https://community.superoffice.com/no/technical/forums/general-forums/announcements/currently-testing-in-sod-api-calls-must-now-use-the-tenants-correct-public-endpoint/
- https://community.superoffice.com/no/technical/forums/general-forums/announcements/breaking-changes-api-urls-for-online-tenants/

This will give some idea of what this additional request returns https://sod.superoffice.com/api/state/Cust12345.

Since this requires additional requests, cache has been added on an 8 hour interval. The cache name chosen matches that of another package I use https://gitlab.com/cadix/web/superoffice-api/-/blob/dev/src/Model.php#L329.

### DIFF
--- a/src/SuperOffice/Provider.php
+++ b/src/SuperOffice/Provider.php
@@ -119,7 +119,7 @@ class Provider extends AbstractProvider
             $api_url = json_decode((string) $response->getBody(), true)['Api'];
 
             if (!$api_url) {
-                throw new \Exception('No API URL received from '. $url);
+                throw new \Exception('No API URL received from '.$url);
             }
 
             return $api_url.'/v1/';
--- a/src/SuperOffice/Provider.php
+++ b/src/SuperOffice/Provider.php
@@ -69,11 +69,7 @@ class Provider extends AbstractProvider
     protected function getUserByToken($token): array
     {
         $response = $this->getHttpClient()->get(
-            sprintf(
-                'https://%s.superoffice.com/%s/api/v1/User/currentPrincipal',
-                $this->getConfig('environment', 'sod'),
-                $this->getConfig('customer_id')
-            ),
+            $this->getBaseApiUrl().'User/currentPrincipal',
             [
                 RequestOptions::HEADERS => [
                     'Accept'        => 'application/json',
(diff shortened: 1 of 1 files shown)


src/SuperOffice/Provider.php
```diff
@@ -105,4 +101,28 @@ public static function additionalConfigKeys(): array
...
         ];
     }
+
+    private function getBaseApiUrl(): string
+    {
+        $cache_time = 60 * 60 * 8; // 8 hours
+        $environment = $this->getConfig('environment', 'sod');
+        $customer_id = $this->getConfig('customer_id');
+
+        return cache()->remember('superoffice-base-url', $cache_time, function () use ($environment, $customer_id) {
```
atymic, line 111:
```suggestion
        return cache()->remember('superoffice-base-url', now()->addHours(8), function () use ($environment, $customer_id) {
```

src/SuperOffice/Provider.php
```diff
@@ -105,4 +101,28 @@ public static function additionalConfigKeys(): array
             'customer_id',
         ];
     }
+
+    private function getBaseApiUrl(): string
+    {
+        $cache_time = 60 * 60 * 8; // 8 hours
+        $environment = $this->getConfig('environment', 'sod');
```
atymic, line 108: Camelcase please

src/SuperOffice/Provider.php
```diff
@@ -105,4 +101,28 @@ public static function additionalConfigKeys(): array
...
+
+        return cache()->remember('superoffice-base-url', $cache_time, function () use ($environment, $customer_id) {
+            $url = sprintf(
+                'https://%s.superoffice.com/api/state/%s',
+                $environment,
+                $customer_id
+            );
+
+            $response = $this->getHttpClient()->get($url);
+            $api_url = json_decode((string) $response->getBody(), true)['Api'];
```
atymic, line 119: Do we need a null coalesce here? I doubt the check below will get hit because if this prop is missing then it will error here.

atymic (changes requested):

atymic (approved):

PR #1012, opened 2023-04-29
Allow for Azure AD B2C custom domains

kurucu:
Azure AD B2C allows for custom domains. As the Azure provider has the url hardcoded, this wasn't possible. This PR attempts to resolve that.


kurucu:
Will stop now - this is covered here `src/AzureADB2C` but not in the docs at https://socialiteproviders.com?

PR #1013, opened 2023-04-29
Azure AD B2C Custom Domains

kurucu:
Azure AD B2C allows for [Custom Domains](https://learn.microsoft.com/en-us/azure/active-directory-b2c/b2clogin#change-identity-provider-redirect-urls), as well as use of b2clogin.com, as the former isolates your cookies and user authentication.

As such, the possible URLs we need to support are as follows:

- https://{your-tenant-name}.b2clogin.com/{your-tenant-id}/oauth2/authresp
- https://{your-tenant-name}.b2clogin.com/{your-tenant-name}.onmicrosoft.com/oauth2/authresp
- https://{your-custom-domain}/{your-tenant-id}/oauth2/authresp
- https://{your-custom-domain}/{your-tenant-name}.onmicrosoft.com/oauth2/authresp


kurucu:
This is ready for review now.

atymic:
@fujie
@lsmith77

Any thoughts on this one?

lsmith77:
I do not have a setup to test this, but it is correct that there is a possibility for custom domains.

fujie:
no objection from me. It seems like that the custom domain url patterns are correct.

PR #1014, opened 2023-05-02
 facebook returned user fields, add nickname

rizkytegar:
**example returned user fields**

```php
Laravel\Socialite\Two\User {#423  // app/Http/Controllers/Module/AuthModule.php:129
  +id: "7677906744xxxxx"
  +nickname: erzedxx
  +name: "Erzedxx"
  +email: null
  +avatar: "https://graph.facebook.com/v3.3/767790674406069/picture?type=normalxx"
  +user: array:2 [
    "name" => "Erzedxxx"
    "id" => "7677906744xxxxx"
  ]
  +attributes: array:7 []
  +token: ""
  +refreshToken: null
  +expiresIn: 5183195xx
  +approvedScopes: array:1 [
    0 => ""
  ]
}
```

### DIFF
--- a/src/Facebook/README.md
+++ b/src/Facebook/README.md
@@ -45,5 +45,6 @@ return Socialite::driver('facebook')->redirect();
 
 - ``id``
 - ``name``
+- ``nickname``
 - ``email``
 - ``avatar``


atymic:
That's in the raw data, but not in the default object?
https://github.com/laravel/socialite/blob/5.x/src/Two/FacebookProvider.php#L124

PR #1018, opened 2023-05-04
Apple - lcobucci/jwt 5.0

### LINKED ISSUE #1015: Apple - Broken since 5.5.0
lcobucci/jwt has removed a number of functions:

- Lcobucci\JWT\Configuration::forUnsecuredSigner()
- Lcobucci\JWT\Signer\Key\InMemory::empty()
- Lcobucci\JWT\Signer\None
- Lcobucci\JWT\Token\Signature::fromEmptyData()

See https://lcobucci-jwt.readthedocs.io/en/stable/upgrading/#removal-of-none-algorithm - however 5.5.0 still makes use of `Lcobucci\JWT\Configuration::forUnsecuredSigner()` here https://github.com/SocialiteProviders/Providers/blob/e5cbb4b0c4dd2e20488915a585eda4cac5e25533/src/Apple/Provider.php#L148 which breaks the functionality of this package after upgrade to lcobucci/jwt 5.0

@hamrak @atymic

hamrak:
Add lcobucci/jwt 5.0 Support with fix #1015 .

### DIFF
--- a/src/Apple/AppleSignerInMemory.php
+++ b/src/Apple/AppleSignerInMemory.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SocialiteProviders\Apple;
+
+use Lcobucci\JWT\Signer\Key;
+
+final class AppleSignerInMemory implements Key
+{
+    public string $contents;
+    public string $passphrase;
+
+    /** @param non-empty-string $contents */
+    private function __construct(string $contents, string $passphrase)
+    {
+        $this->passphrase = $passphrase;
+        $this->contents = $contents;
+    }
+
+    /** @param non-empty-string $contents */
+    public static function plainText(string $contents, string $passphrase = ''): self
+    {
+        return new self($contents, $passphrase);
+    }
+
+    public function contents(): string
+    {
+        return $this->contents;
+    }
+
+    public function passphrase(): string
+    {
+        return $this->passphrase;
+    }
+}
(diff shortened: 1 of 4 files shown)


atymic:
@golsson  could you give this a review/test please?
Thanks!

golsson:
Doesn't work for me I'm afraid:
```[2023-05-04 15:30:01] local.ERROR: Key cannot be empty {"exception":"[object] (Lcobucci\\JWT\\Signer\\InvalidKeyProvided(code: 0): Key cannot be empty at /Users/gustav/projects/blah/dev/blah/vendor/lcobucci/jwt/src/Signer/InvalidKeyProvided.php:39)

hamrak:
I'm sorry, i forgot to push the last commit. Check it now.

golsson:
It works now, however I can't comment on the implementation with regards to security etc. since I'm not that familiar with the concepts involved.
One thing I noticed however is the phpdoc for the `plainText()` "constructor", and the `__construct`s in both AppleSignerInMemory.php and it's parent is `/** @param non-empty-string $contents */` but and empty string is hardcoded as input in Provider.php:149? For obvious reasons this doesn't cause any runtime issues but...?
Also on a side note the `getTokenField($code)` method can be removed since `grant_type` is already added in the parent. Let me know if you want me to provide a PR for that.

hamrak:
I will be happy, thank you @golsson .

golsson:
See https://github.com/SocialiteProviders/Providers/pull/1020 regarding `getTokenField($code)`

atymic (approved):

PR #1019, opened 2023-05-04
Apple - new option for using private key instead secret key.

hamrak:
Rather than static client secrets, Apple requires that you derive a client secret yourself from your private key. They use the JWT standard for this, using an elliptic curve algorithm with a P-256 curve and SHA256 hash. In other words, they use the ES256 JWT algorithm.

This PR add new configuration options that allow you to use private key instead of manually generating secret key every 6 months.

> 'key_id' => env('APPLE_KEY_ID'),
> 'team_id' => env('APPLE_TEAM_ID'),
> 'private_key' => env('APPLE_PRIVATE_KEY'), // Must be absolute path, e.g. /var/www/cert/AuthKey_XYZ.p8
> 'passphrase' => env('APPLE_PASSPHRASE'), // Set if your key have a passphrase.
> 'signer' => env('APPLE_SIGNER'), // Signer used for Configuration::forSymmetricSigner().

Need to merge #1018 first.


atymic:
Could we get a docs like and a rebase please :)

hamrak:
I don't know what you mean exactly. 
Official Apple documentation for creating secret token is here: https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens#3262048

I also put some description in README.md.

Can you tell me exactly what I should add? :)

src/Apple/Provider.php
```diff
@@ -170,6 +220,25 @@ public static function verify($jwt)
...
+     * @return bool
+     *
+     * @see https://appleid.apple.com/auth/keys
+     */
+    public static function verify($jwt)
+    {
+        return (new self(
+            new Request(),
+            config('services.apple.client_id'),
+            config('services.apple.client_secret'),
```
atymic, line 237: Will this method still work if the secret is dervied?

hamrak: Yes it will work. I had to do that (create new object) because this function is called statically (don't know why) and it was necessary to call other non static functions.

pricop:
This seems to be inspired from https://bannister.me/blog/generating-a-client-secret-for-sign-in-with-apple-on-each-request

Perhaps add support for private keys that are stored in the database / string format too?

My proposal would be something like:

```
$this->privateKey = $private_key_path; // Support for plain text private keys

if (!empty($private_key_path) && file_exists($private_key_path)) {
    $this->privateKey = file_get_contents($private_key_path);
}
```

This will ensure backwards compatibility for those of us who are already using this integration.

Thank you.

hamrak:
> pricop

Done ;)

pricop:
Yes, it should be ready for review/merge. I've been using this successfully into production on multiple websites. This would be an awesome addition.

hamrak:
Yes, i use it in production too.

src/Apple/Provider.php
```diff
@@ -101,12 +116,53 @@ public function getAccessTokenResponse($code)
...
+            $this->getJwtConfig(); // Generate Client Secret from private key if not set.
+        }
+
+        return $this->clientSecret;
+    }
+
+    protected function getJwtConfig()
+    {
+        if (!$this->jwtConfig) {
+            $private_key_path = config('services.apple.private_key', '');
```
atymic, line 137: can we access this via the standard method, ie add to 

```
    {
        return ['private_key', 'otherkey...'];
    }
```

```
$this->getConfig('private_key', '')
```

src/Apple/Provider.php
```diff
@@ -101,12 +116,53 @@ public function getAccessTokenResponse($code)
...
+                $signer = !empty($private_key_path) ? \Lcobucci\JWT\Signer\Ecdsa\Sha256::class : AppleSignerNone::class;
+            }
+            
+            $this->privateKey = $private_key_path; // Support for plain text private keys
+
+            if (!empty($private_key_path) && file_exists($private_key_path)) {
+                $this->privateKey = file_get_contents($private_key_path);
+            }
+
+            $this->jwtConfig = Configuration::forSymmetricSigner(
```
atymic, line 151: Just to confirm, there is no bc break here, existing integrations using the secret key will continiue to work fine?

atymic (changes requested):
One small change and a comment.

atymic:
@hamrak just a quick thing re; config and a question, then I will merge and tag new vers :)

src/Apple/Provider.php
hamrak: yes, if you pass the path to the private key file, it will get the contents of the private key file, otherwise you put the private key contents by default.

src/Apple/Provider.php
hamrak: Yes, we can. Changed.

src/Apple/Provider.php
```diff
@@ -101,12 +116,53 @@ public function getAccessTokenResponse($code)
...
+                new $signer(),
+                AppleSignerInMemory::plainText($this->privateKey, $private_key_passphrase)
+            );
+
+            if (!empty($this->privateKey)) {
+                $appleToken = new AppleToken($this->getJwtConfig());
+                $this->clientSecret = $appleToken->generate();
+                config()->set('services.apple.client_secret', $this->clientSecret);
```
atymic, line 159: Might need to set this to a variable on the class with above switch to `$this->getConfig`

src/Apple/Provider.php
```diff
@@ -168,6 +220,25 @@ public static function verify($jwt)
...
+     *
+     * @return bool
+     *
+     * @see https://appleid.apple.com/auth/keys
+     */
+    public static function verify($jwt)
+    {
+        return (new self(
+            new Request(),
+            config('services.apple.client_id'),
```
atymic, line 236:
```suggestion
            $this->getConfig('client_id', '')
```

And other calls to `config()`

atymic (changes requested):

src/Apple/Provider.php
hamrak: You can call $this-> in static function, or ?

atymic: oh my bad, didn't realise it was static. Is that why it overrides the global config above?

atymic (approved):

src/Apple/Provider.php
hamrak: Why do you think so?

atymic:
cc @hamrak sorry can you resolve conflicts and we can get this in

hamrak:
> cc @hamrak sorry can you resolve conflicts and we can get this in

Done.

src/Apple/Provider.php
hamrak: > oh my bad, didn't realise it was static. Is that why it overrides the global config above?

yes

hamrak:
@atymic Is there a problem in my PR? or does something need to be changed? so that it can finally merge? :)

src/Apple/Provider.php
SvenRtbg: Do you consider writing to the configuration a good idea? I would probably not expect this provider to overwrite the client secret that was defined with something else.

src/Apple/Provider.php
```diff
@@ -88,12 +106,53 @@ public function getAccessTokenResponse($code)
...
+                $this->privateKey = $private_key_path; // Support for plain text private keys
+            }
+
+            $this->jwtConfig = Configuration::forSymmetricSigner(
+                new $signer(),
+                AppleSignerInMemory::plainText($this->privateKey, $private_key_passphrase)
+            );
+
+            if (!empty($this->privateKey)) {
+                $appleToken = new AppleToken($this->getJwtConfig());
```
SvenRtbg, line 147: Not too happy about this recursive call. The config is already created a few lines earlier, and it does seem like it would work if used directly.

manhtruongwang:
This merged cause every implementation before 5.7.0 broke. Here is the error:
`error:1E08010C:DECODER routines::unsupported`

atymic:
@manhtruongwang Feel free to PR a fix if you found an issue. I doubt it caused every app to break otherwise someone  would have commented.

veneliniliev:
@manhtruongwang Did you find a solution to the problem?

PR #1021, opened 2023-05-05
Fix: set access token properly when fetching user with apple

sunchayn:
The `token` of the `\Laravel\Socialite\Two\User` class is meant to hold the access token.

The `id_token` acquired from Apple authorization requests is not a valid access token.

Also, I've clarified the intent of `getUserByToken` which unlike its parent it expects an identity token instead of an access token.

---

I would go and deprecate `userByIdentityToken` in favor of `getUserByToken`, but unsure what are the other opinions about this.

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -97,7 +97,11 @@ class Provider extends AbstractProvider
     }
 
     /**
-     * {@inheritdoc}
+     * Get the raw user for the given identity token.
+     *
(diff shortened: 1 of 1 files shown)


atymic:
Is there any breaking changes in this PR?

atymic:
Also rebase please :)

atymic:
@hamrak thoughts?

PR #1031, opened 2023-05-29
Update README.md and Provider.php with simplified configuration

crThiago:
The following changes have been made to simplified the configuration in the Provider.php file:

README.md:

    Removed the 'environment' key from the 'govbr' configuration array.

Provider.php:

    Simplified the getBaseUrlForEnvironment() method to use the 'app.env' configuration value instead of the 'environment' key.
    Replaced the match statement with a conditional expression to determine the base URL based on the 'app.env' value.

### DIFF
--- a/src/GovBR/Provider.php
+++ b/src/GovBR/Provider.php
@@ -131,12 +131,6 @@ class Provider extends AbstractProvider implements ProviderInterface
      */
     protected function getBaseUrlForEnvironment(): string
     {
-        $environment = $this->getConfig('environment', 'production');
-
-        return match ($environment) {
-            'staging'    => $this->stagingUrl,
-            'production' => $this->productionUrl,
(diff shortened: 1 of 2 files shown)


src/GovBR/Provider.php
```diff
@@ -131,12 +131,6 @@ public static function additionalConfigKeys()
...
     protected function getBaseUrlForEnvironment(): string
     {
-        $environment = $this->getConfig('environment', 'production');
-
-        return match ($environment) {
-            'staging'    => $this->stagingUrl,
-            'production' => $this->productionUrl,
-            default      => throw new RuntimeException("Invalid environment '{$environment}' selected for GovBR provider."),
-        };
+        return config('app.env') === 'production' ? $this->productionUrl : $this->stagingUrl;
```
atymic, line 134: This would be a breaking change right? This might cause apps to not function correctly, so I don't think this is a good idea

PR #1035, opened 2023-06-08
Fix for bug when user is not in organization 

CosminBd:
I've encountered an issue when you put in config` 'include_tenant_info' => true` If the user is not part of an organisation and you try to retrieve it's tenant, then Microsoft API returns a 500 status code error (for some reason), and this breaks execution.

My fix wraps the API call in a try/catch block and if a Client exception occurs, it sets the tenant to null.


atymic (approved):

atymic:
Could you fix the style issues please, thanks.

CosminBd:
done, thanks for pointing out

PR #1038, opened 2023-06-21
Update Provider.php to display Default avatars

kr3t4r:
Default avatars aren't displayed currently , to add this i added this change . Discord uses the last digit of the discriminator to get the avatar .


atymic:
This seems like a fallback avatar. Should this not be handled in userland code?

kr3t4r:
No it is the

> This seems like a fallback avatar. Should this not be handled in userland code?

No it is the same avatar as the user, if he didnt upload a custom one.

PR #1042, opened 2023-06-28
On nickname discriminator is now displayed only if the user has assigned one

Havenstd06:
With discord's new naming system, only displays the discriminator if the user has one.


atymic:
Could you link to the API docs change for this please?

atymic (approved):

PR #1046, opened 2023-07-16
Add cache keying by metadata configuration

### LINKED ISSUE #1037: Saml2: Collisions in metadata caching
https://github.com/SocialiteProviders/Providers/blob/6d291f4e33dd04632d92d43e4428e352dd7a721f/src/Saml2/Provider.php#LL303C53-L303C53

This method caches values using a hardcoded key irrespective of the URL being fetched. When the URL changes, or one e.g. uses multiple SAML2 configurations, this can cause collisions.

Could it be possible to make these cache keys scoped to the URL? Or otherwise, make it so that the key can be overridden in a child Provider class?

27pchrisl:
Fixes #1037

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -96,11 +96,6 @@ class Provider extends AbstractProvider implements SocialiteProvider
      */
     protected $config;
 
-    public const CACHE_NAMESPACE = 'socialite_saml2';
-    public const METADATA_CACHE_KEY = self::CACHE_NAMESPACE.'_metadata';
-    public const METADATA_CACHE_KEY_TTL = self::METADATA_CACHE_KEY.'_ttl';
-    public const ID_CACHE_PREFIX = self::CACHE_NAMESPACE.'_id_';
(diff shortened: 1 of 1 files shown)


src/Saml2/Provider.php
```diff
@@ -736,4 +731,10 @@ protected function mapUserToObject(array $user)
     {
         throw new NotSupportedException();
     }
+
+    protected function cacheKey(string $key): string
+    {
+        $hash = md5($this->getConfig('acs') ?: $this->getConfig('metadata'));
+        return "socialite_saml2_{$hash}_{$key}";
```
atymic, line 738: could we use `sprintf` here please?

atymic (commented):
LGTM, one minor comment.

src/Saml2/Provider.php
27pchrisl: Done, and fixed the other issue the linter found

atymic:
Thanks!

PR #1048, opened 2023-07-18
Update firebase/php-jwt to latest version (6.8) 

cwp-jamesh:
The current version set in the composer file for AzureADB2C is incompatible with the current version of Laravel passport (requires 6.4 or higher).

### DIFF
--- a/src/AzureADB2C/Provider.php
+++ b/src/AzureADB2C/Provider.php
@@ -186,7 +186,7 @@ class Provider extends AbstractProvider
             }
 
             // signature validation and return claims
-            return (array) JWT::decode($idToken, JWK::parseKeySet($this->getJWTKeys(), $this->getConfig('default_algorithm')), $this->getOpenIdConfiguration()->id_token_signing_alg_values_supported);
+            return (array) JWT::decode($idToken, JWK::parseKeySet($this->getJWTKeys(), $this->getConfig('default_algorithm')));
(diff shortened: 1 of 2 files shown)


cwp-sheal (approved):
LGTM

src/AzureADB2C/Provider.php
```diff
@@ -186,7 +186,7 @@ private function validateIdToken($idToken)
             }
 
             // signature validation and return claims
-            return (array) JWT::decode($idToken, JWK::parseKeySet($this->getJWTKeys(), $this->getConfig('default_algorithm')), $this->getOpenIdConfiguration()->id_token_signing_alg_values_supported);
+            return (array) JWT::decode($idToken, JWK::parseKeySet($this->getJWTKeys(), $this->getConfig('default_algorithm')));
```
atymic, line 189: Can you link to the change as to why the supported algs were removed?

cwp-jamesh: Sure, it was a breaking change of version 6.0.0: https://github.com/firebase/php-jwt/releases/tag/v6.0.0

atymic (approved):

src/AzureADB2C/Provider.php
cwp-jamesh: Thanks a lot @atymic 👍

PR #1055, opened 2023-08-21
Bump to PHP 8.1

lucasmichot:
This PR bumps the minimum version requirement to PHP 8.1
As PHP 8.0 will be EOL in a few months.

Cf. https://www.php.net/supported-versions.php

---

The GA PHP versions are updated accordingly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,14 +11,14 @@ jobs:
       matrix:
         validate: [ false ]
         fail-fast: [ false ]
-        php-versions: ['8.0', '8.1' ]
+        php-versions: [ 8.1, 8.2 ]
         exclude:
-          - php-versions: '8.0'
-          - php-versions: '8.1'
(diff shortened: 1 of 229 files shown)


atymic:
TBH i'm not a fan of removing supported versions just because they are EOL. AFAIK there's nothing we need that's not available in `8.0`, and despite it being EOL it's still used (has a 10% share)
https://packagist.org/php-statistics

lucasmichot:
@atymic Should we keep on supporting PHP versions 7.* then?

atymic:
AFAIK we dropped 7.x support? Maybe revisit once 8.0 is EOL?

lucasmichot:
> AFAIK we dropped 7.x support? Maybe revisit once 8.0 is EOL?

The support is actually not totally dropped, I have created https://github.com/SocialiteProviders/Providers/pull/1060 for that

Let's revisit this PR later

PR #1063, opened 2023-08-23
Normalize all composer.json files

lucasmichot:
This PR normalizes all `composer.json` files

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "php": "^8.0",
         "ext-json": "*",
         "league/oauth1-client": "^1.9",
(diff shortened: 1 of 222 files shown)


atymic (commented):
While we're doing this, worth bumping manager requirement to 4.4.0?

lucasmichot:
> While we're doing this, worth bumping manager requirement to 4.4.0?

Good idea, see https://github.com/SocialiteProviders/Providers/pull/1063/commits/758c183bba2c279e5eaa828cad0adcaaec6c00d3

atymic (approved):

PR #1064, opened 2023-08-23
Add a coding standards GA to fix code with Laravel Pint.

lucasmichot:
This PR adds a coding standards GA to fix code with Laravel [Pint](https://laravel.com/docs/10.x/pint).

It also fixes the code to match those coding standards

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,0 +1,44 @@
+name: Fix code styling
+
+on:
+  workflow_dispatch:
+    inputs:
+      php:
+        default: "8.1"
+        description: The PHP version to use
+        type: string
+      message:
+        default: Fix code styling
+        description: The commit message to use
+        type: string
+      fix:
+        default: true
+        description: Whether to fix the code styling
+        type: boolean
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v3
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ inputs.php }}
+          extensions: json, dom, curl, libxml, mbstring
+          coverage: none
+
+      - name: Install Pint
+        run: composer global require laravel/pint
+
+      - name: Run Pint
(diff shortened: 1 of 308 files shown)


src/Acclaim/Provider.php
```diff
@@ -51,11 +51,11 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'       => $user['data']['id'],
+            'id' => $user['data']['id'],
             'nickname' => $user['data']['first_name'],
-            'name'     => $user['data']['first_name'],
```
atymic, line 56: Could we leave the align arrays rule on? changes such as these just pollute the commit history imo.

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,44 @@
+name: Fix code styling
+
+on:
+  workflow_dispatch:
```
atymic, line 4: Should this also not run on commit to the master branch as well?

lucasmichot: I would probably stick to a dispatch, to mimic 
https://github.com/laravel/.github/blob/main/.github/workflows/coding-standards.yml

src/Acclaim/Provider.php
lucasmichot: Sure 👍

.github/workflows/coding-standards.yml
atymic: I mean, just the lint (not the fix)

atymic:
Could you resolve the conflicts?

lucasmichot:
> Could you resolve the conflicts?

Conflicts resolved

.github/workflows/coding-standards.yml
lucasmichot: I see; in any case, if lint should format for workflow_dispatch and lint on push, we need 2 different workflows files.
I propose to merge this one, and I'll make a PR for the linting part on push into master

WDYT ?

atymic: Can we have a lint workflow for PRs as well please?

PR #1067, opened 2023-08-25
Add SP initiated logoutRequest, add RelayState in messageContext

### LINKED ISSUE #978: SAML2, logout throws "unknown binding type"
## What:

When implementing the single sign out feature on the service provider as indicated [here](https://socialiteproviders.com/Saml2/#installation-basic-usage) the application is throwing an "Unknown binding type" error instead of generating a SAML2 logout request and redirect.

## When:
When implementing the single sign out

## How:
Following the instructions in the SAML2 provider documentation, I have added the following in my `routes/web.php` file:
```php
Route::get('/saml2/logout', [\App\Http\Controllers\Logout::class, 'logout']);
```
My Logout controller has a single function:
```php
    public function logout()
    {
        /** @var \SocialiteProviders\Saml2\Provider $response */
        $response = Socialite::driver(self::DRIVER);
        return $response->logoutResponse();
    }
```

For testing purposes I omitted logging the user out locally.

## What is expected:
The get call should result in a redirect to the single sign out URL as configured in my SAML configuration with the appropriately configured LogOut asssertion

## What is wrong:
I am not quite sure. I have tracked the issue down to the `BindingFactory.php` `processGet()` function. It is looking for a request query that does not exists - the url given is a simple GET call to a URL with no additional data. The documentation does not give any additional setup instructions beyond some session cookie settings so nothing is every going to be given additionally. The URL provided by the application is what should start the single sign out request to the identity provider.

I may be missing something but I am not sure.

aleksanm:
I created this PR because today I needed SAML Single Logout from Service Providers (SP) to Microsoft ADFS, so I added logoutRequest method to Providers class and added chance to specify RelayState in SAML messageContext.
Tested with MS Azure AD, both directions: SLO from IDP to SP and SP to IDP. Worked.

Closes #978

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -31,6 +31,7 @@ use LightSaml\Helper;
 use LightSaml\Model\Assertion\Assertion;
 use LightSaml\Model\Assertion\AttributeStatement;
 use LightSaml\Model\Assertion\Issuer;
+use LightSaml\Model\Assertion\NameID;
 use LightSaml\Model\Context\DeserializationContext;
 use LightSaml\Model\Context\SerializationContext;
 use LightSaml\Model\Metadata\AssertionConsumerService;
@@ -43,6 +44,7 @@ use LightSaml\Model\Metadata\Organization;
 use LightSaml\Model\Metadata\SingleLogoutService;
 use LightSaml\Model\Metadata\SpSsoDescriptor;
 use LightSaml\Model\Protocol\AuthnRequest;
+use LightSaml\Model\Protocol\LogoutRequest;
 use LightSaml\Model\Protocol\LogoutResponse;
 use LightSaml\Model\Protocol\NameIDPolicy;
 use LightSaml\Model\Protocol\SamlMessage;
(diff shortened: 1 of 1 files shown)


atymic:
@27pchrisl

atymic:
Looks like it's back up again @27pchrisl if you could review when you have a chance :)

src/Saml2/Provider.php
```diff
@@ -241,6 +260,11 @@ protected function sendMessage(SamlMessage $message, string $bindingType): HttpF
         $messageContext = new MessageContext();
         $messageContext->setMessage($message);
 
+        if (! empty(request()->filled('RelayState'))) {
```
27pchrisl, line 263: This would be better as:

```php
        if ($this->messageContext->getMessage() instanceof SamlMessage) {
            $messageContext->getMessage()->setRelayState($this->messageContext->getMessage()->getRelayState());
        }
```

So that it's referring to the same inbound messagecontext as the rest of the code.

27pchrisl (changes requested):
Thanks @aleksanm for your contribution! I've requested one small change.

27pchrisl:
@atymic looks good to me!

PR #1079, opened 2023-09-05
Fix docblocks

lucasmichot:
This PR fix some docblocks

### DIFF
--- a/src/Flexkids/Provider.php
+++ b/src/Flexkids/Provider.php
@@ -125,13 +125,7 @@ class Provider extends AbstractProvider
         return ['resource', 'apiuser', 'authurl', 'server'];
     }
 
-    /**
-     * Set the token.
-     *
-     * @param  string  $idToken
-     * @return $this
-     */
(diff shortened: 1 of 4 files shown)


src/Flexkids/Provider.php
```diff
@@ -126,8 +126,10 @@ public static function additionalConfigKeys()
...
     /**
-     * @param $idToken
-     * @return Provider
+     * Set the token.
+     *
+     * @param  string  $idToken
+     * @return $this
      */
     private function setIdToken($idToken)
     {
```
atymic, line 135: Could we just add typres and remove the docblock?

atymic: (same with others)

lucasmichot: cf. https://github.com/SocialiteProviders/Providers/pull/1079/commits/246cf0fbe08db05ff334c06202136d14081dd03a

atymic (approved):

PR #1081, opened 2023-09-05
Authentik: add refreshToken method

RixzZ:
Added ability to refreshToken to the Authentik provider based on https://github.com/SocialiteProviders/Providers/pull/1022/commits/07f607f9c0e73b18324faa609ae6aa9bb4e0062f.

### DIFF
--- a/src/Authentik/Provider.php
+++ b/src/Authentik/Provider.php
@@ -39,7 +39,7 @@ class Provider extends AbstractProvider
      */
     protected function getAuthUrl($state)
     {
-        return $this->buildAuthUrlFromBase($this->getBaseUrl() . '/application/o/authorize/', $state);
+        return $this->buildAuthUrlFromBase($this->getBaseUrl().'/application/o/authorize/', $state);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Authentik/Provider.php
```diff
@@ -56,17 +56,32 @@ protected function getTokenUrl()
...
             ]
         );
 
         return json_decode((string) $response->getBody(), true);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    protected function refreshToken($refreshToken)
```
atymic, line 73: could we type this to string?

atymic:
@lucasmichot I don't think linting is running correctly here?

src/Authentik/Provider.php
RixzZ: I guess so. I didn't type it since the rest of the provider isn't.

Should I add the type then?

lucasmichot:
> Style fixing

Indeed it should after commit https://github.com/SocialiteProviders/Providers/pull/1081/commits/5e0ead4504b53378b085b57b6ee1ad55bbe3156f, no clue why it does not
I need a bit of time to find out

atymic:
> > Style fixing
> 
> I does not run after every commit (`workflow_disptach`)

Can we have at least a lint on PRs?

lucasmichot:
sure

cf. my comment https://github.com/SocialiteProviders/Manager/pull/221#issuecomment-1700790648 too

PR #1082, opened 2023-09-08
Spelling & Documentation Fix

shealavington:


atymic:
Can you explain your changes with the docs please?

shealavington:
> Can you explain your changes with the docs please?

The JWK function used in this package to authenticate does not work unless an algorithm is specified, it throws an error that "alg has not been set".

By default, the package was set to null/undefined causing the package to not work. Instead, I have changed the package default to the recommended auth default. This of course can still be overridden using the env.

atymic (approved):

PR #1093, opened 2023-10-11
Normalize the case of the identifiers names

lucasmichot:
This PR normalizes the case of the identifiers names

### DIFF
--- a/src/MakerLog/Provider.php
+++ b/src/MakerLog/Provider.php
@@ -8,7 +8,7 @@ use SocialiteProviders\Manager\OAuth2\User;
 
 class Provider extends AbstractProvider
 {
-    public const IDENTIFIER = 'makerlog';
+    public const IDENTIFIER = 'MAKERLOG';
 
     /**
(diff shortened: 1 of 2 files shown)


atymic:
@lucasmichot is this going to need a new release?

lucasmichot:
Ye

> @lucasmichot is this going to need a new release?

Yes @atymic

lucasmichot:
> Sorry, I mean will it be a breaking change? @lucasmichot

Yes it does

atymic:
> > Sorry, I mean will it be a breaking change? @lucasmichot
> 
> Yes it does

Do we need to add update notes for existing apps, then?

lucasmichot:
No we don't, the identifier is anyway used as lowercase in the `\SocialiteProviders\Manager\Helpers\ConfigRetriever`

atymic:
> ConfigRetriever

Looks like this pulls from the `extendSocialite` method, not the identifier one. So this is fine.

PR #1095, opened 2023-10-11
Minimize the number of spaces used for array alignment

lucasmichot:
This PR minimizes the number of spaces used for array alignment


atymic:
TBH I don't think this is worth polluting the change history for.

PR #1098, opened 2023-10-13
AdmitAd: use space as the scope separator

josephkerkhof:
According to the [AdmitAd scope documentation](https://developers.admitad.com/hc/en-us/articles/7930273820689-Help-on-access-settings):

> it is required to send `scope` parameter containing names of rights separated by space.

This change should fix this up so it is possible to use the Laravel `->setScopes()` function.


lucasmichot (approved):
GTM @atymic

PR #1099, opened 2023-10-17
fix: apple JWT token expired

zencodex:
the issue here: new LooseValidAt(SystemClock::fromSystemTimezone())

when $leeway is null, here `new DateInterval('PT0S');` will cause JWT token is expired.

```
    public function __construct(Clock $clock, ?DateInterval $leeway = null)
    {
        $this->clock  = $clock;
        $this->leeway = $this->guardLeeway($leeway);
    }

    private function guardLeeway(?DateInterval $leeway): DateInterval
    {
        if ($leeway === null) {
            return new DateInterval('PT0S');
        }

        if ($leeway->invert === 1) {
            throw LeewayCannotBeNegative::create();
        }

        return $leeway;
    }
 ```
 
 print log for JWT exp: 
 
 ```
  'exp' =>
  \DateTimeImmutable::__set_state(array(
     'date' => '2023-10-17 10:21:21.000000',
     'timezone_type' => 1,
     'timezone' => '+00:00',
  )),
  'iat' =>
  \DateTimeImmutable::__set_state(array(
     'date' => '2023-10-16 10:21:21.000000',
     'timezone_type' => 1,
     'timezone' => '+00:00',
  )),
  ```
  so it would be better `PT6H` (6 hours) as $leeway. I have tested with my app, and it would be successful.

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -153,7 +153,7 @@ class Provider extends AbstractProvider
             $constraints = [
                 new SignedWith(new Sha256(), AppleSignerInMemory::plainText($publicKey['key'])),
                 new IssuedBy(self::URL),
-                new LooseValidAt(SystemClock::fromSystemTimezone()),
+                new LooseValidAt(SystemClock::fromSystemTimezone(), new \DateInterval('PT6H')),
             ];
 
             try {

### CODE REVIEW

**Comment by atymic** [comment]
I don't think 6 hours of leeway is a good idea. Maybe like 5 minutes, maximum?

**Comment by zencodex** [comment]
screen

I tested again with 5 minutes, but also met this issue, see above screen.gif.
I am not sure the regular rule, seems that several hours would be safe.

**Comment by atymic** [comment]
What is your system time? The issue is probably there.

**Comment by zencodex** [comment]
I think I found the root cause, for the oauth ios app and api MUST handle with the right flow.

```
// api code
$token = $request->get('token');
$socialDriver = Socialite::driver($provider);
$socialUser = $socialDriver->userFromToken($token);
```

jwt token is from app client, if we don't revoke the token, the app client would send the same jwt token next time, the token may be expired.

so I should revoke after the token used. the end api code like following:

```
$token = $request->get('token');
$socialDriver = Socialite::driver($provider);
$socialUser = $socialDriver->userFromToken($token);

// add revoke here.
$socialDriver->revokeToken($token);
```

I will trace the issue next several days, I will close the pull request after this issue cannot be reproduced. thanks a lot for the great project.

zencodex:
I have confirmed, the ios client cached a issued token as session long time, ios client should clean the session and get the new access token. we needn't to fix, thanks.

atymic:
Glad you solved it, worth adding a note to the docs?

PR #1126, opened 2023-12-06
Fixes for Apple signature

turbo124:
the refreshToken method signature does not match AbstractProvider which will result in a fatal error.

this PR sets the signature to match.


atymic (approved):

rcerljenko:
@atymic could you trigger a new release maybe?

PR #1129, opened 2023-12-12
Update OneLogin `Provider@getRefreshTokenResponse()`

cosmastech:
Similar to https://github.com/SocialiteProviders/Providers/pull/1128, this fixes the signature for OneLogin (created by upstream change method added in https://github.com/laravel/socialite/pull/675)


cosmastech:
From my quick scan, this is the last provider that needs the signature change.

atymic (approved):

PR #1134, opened 2023-12-27
Add Indeed OAuth Provider

sahapranta:
Hi,

I am requesting to add indeed OAuth Provider, Here is the doc

https://docs.indeed.com/dev/docs/log-in-with-indeed

https://docs.indeed.com/dev/docs/upgrade-oauth

### DIFF
--- a/src/Indeed/Provider.php
+++ b/src/Indeed/Provider.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace SocialiteProviders\Indeed;
+
+use GuzzleHttp\RequestOptions;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    public const IDENTIFIER = 'INDEED';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    protected $scopes = ['email', 'offline_access'];
+
+    /**
+     * @inheritDoc
+     */
+    protected function getAuthUrl($state)
+    {
+        return $this->buildAuthUrlFromBase('https://secure.indeed.com/oauth/v2/authorize', $state);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function getTokenUrl()
+    {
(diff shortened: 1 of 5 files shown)


src/Indeed/Provider.php
```diff
@@ -0,0 +1,57 @@
...
+class Provider extends AbstractProvider
+{
+    public const IDENTIFIER = 'INDEED';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    protected $scopes = ['email', 'offline_access'];
```
atymic, line 18: Do we need `offline_access` token by default? I assume we can omit this in the provider

atymic (changes requested):
See comments

src/Indeed/Provider.php
sahapranta: Hi @atymic, I don't know why but they require to have `offline_access` included when requesting for `email`

[https://docs.indeed.com/dev/docs/authentication-overview#scopes](https://docs.indeed.com/dev/docs/authentication-overview#scopes:~:text=Incremental%20authorization%20enables%20apps%20to%20request%20only%20the%20scopes%20that%20they%20need%2C%20and%20requires%20the%20offline_access%2C%20which%20maintains%20the%20permissions%20that%20your%20app%20has.%20For%20example%2C%20if%20your%20app%20requests%20the%20email%20scope%2C%20your%20app%20must%20also%20request%20the%20offline_access%20scope.)

Here is a screenshot from the above url.

image

src/Indeed/README.md
```diff
@@ -0,0 +1,42 @@
...
+```php
+protected $listen = [
+    \SocialiteProviders\Manager\SocialiteWasCalled::class => [
+        // ... other providers
+        \SocialiteProviders\Indeed\IndeedExtendSocialite::class.'@handle',
+    ],
+];
+```
+
+### Usage
```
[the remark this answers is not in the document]
sahapranta: Thanks for the mention & Updated the docs.

src/Indeed/Provider.php
atymic: Ahh ok.

atymic:
Please PR to website repo

PR #1135, opened 2023-12-29
Update in accordance with discord changes

WissemBad:
The **DisplayName**, introduced this year by Discord, can now be found alongside the **Username** and **Discriminator**.

> You'll find all the information about the fields currently being returned on the [Discord Developer Portal - Users Resource](https://discord.com/developers/docs/resources/user).

## New returned fields :
-   `id` : The user's id (snowflake).
-   `nickname` : The user's name with discriminator (name#0000).
-   `username` : The user's username (name).
-   `discriminator` : The user's Discord-tag (0000).
-   `global_name` : The user's display name, if it is set (Amazing Name).
-   `email` : The user's email (name@example.com).
-   `avatar` : The user's avatar URL.

### DIFF
--- a/src/Discord/Provider.php
+++ b/src/Discord/Provider.php
@@ -116,8 +116,9 @@ class Provider extends AbstractProvider
     {
         return (new User())->setRaw($user)->map([
             'id'            => $user['id'],
+            'nickname'      => $user['username'].($user['discriminator'] !== '0' ? '#'.$user['discriminator'] : '')
             'username'      => $user['username'],
-            'discriminator' => $user['discriminator'],
+            'discriminator' => $user['discriminator'] ?? null,
             'global_name'   => $user['global_name'],
             'email'         => $user['email'] ?? null,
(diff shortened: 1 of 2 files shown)


src/Discord/Provider.php
```diff
@@ -115,11 +115,13 @@ protected function formatAvatar(array $user)
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
-            'id'       => $user['id'],
-            'nickname' => $user['username'].($user['discriminator'] !== '0' ? '#'.$user['discriminator'] : ''),
-            'name'     => $user['username'],
```
atymic, line 120: Removing the name field here would be a breaking change.

atymic: Happy to dd the `username` field in addition,  but they can always use `getRaw` to get the other fields.

atymic (changes requested):

atymic:
ping @WissemBad 
Will close this if no update in next few weeks.

PR #1144, opened 2024-02-01
bump min PHP version to 8.2

matt-moellering:
For Laravel 11, which requires 8.2


atymic (changes requested):
No need to drop support for older versions, it will work fine.

PR #1148, opened 2024-02-14
Use AR as default country for MercadoLibre.

lucasmichot:
This PR ensures that the default country code to be used is `AR`

### DIFF
--- a/src/MercadoLibre/Provider.php
+++ b/src/MercadoLibre/Provider.php
@@ -13,27 +13,6 @@ class Provider extends AbstractProvider
 {
     public const IDENTIFIER = 'MERCADOLIBRE';
 
-    public const DOMAIN = [
-        'AR' => 'https://auth.mercadolibre.com.ar',
-        'BO' => 'https://auth.mercadolibre.com.bo',
-        'BR' => 'https://auth.mercadolivre.com.br',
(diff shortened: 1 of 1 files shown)


src/MercadoLibre/Provider.php
```diff
@@ -57,9 +36,30 @@ public static function additionalConfigKeys()
      */
     protected function getAuthUrl($state)
     {
-        $url = self::DOMAIN[$this->getConfig('country')] ?? 'https://auth.mercadolibre.com.ar';
+        $domains = [
```
atymic, line 39: is there a reason for moving this out of the const?
Isn't this the same end result

lucasmichot: Not really, I revered that

src/MercadoLibre/Provider.php
```diff
@@ -57,9 +57,9 @@ public static function additionalConfigKeys()
      */
     protected function getAuthUrl($state)
     {
-        $url = self::DOMAIN[$this->getConfig('country')] ?? 'https://auth.mercadolibre.com.ar';
```
atymic, line 60: I actually think the original is better, as if they pass an invalid country code it won't cause an error. What do you think?

PR #1167, opened 2024-03-04
Add refreshToken for HubSpot

dansleboby:
Add refreshToken method for Hubspot

Inspire by:
https://github.com/SocialiteProviders/Apple/blob/master/Provider.php#L289-L299

### DIFF
--- a/src/HubSpot/Provider.php
+++ b/src/HubSpot/Provider.php
@@ -70,13 +70,13 @@ class Provider extends AbstractProvider
      * @see https://developers.hubspot.com/docs/api/oauth-quickstart-guide#refreshing_tokens
      *
      * @param string $refreshToken
-     * @return ResponseInterface
+     * @return array
      *
      * @throws \GuzzleHttp\Exception\GuzzleException
      */
     public function refreshToken($refreshToken): ResponseInterface
     {
-        return $this->getHttpClient()->post($this->getTokenUrl(), [
+        $response = $this->getHttpClient()->post($this->getTokenUrl(), [
             RequestOptions::HEADERS => [
                 'Accept' => 'application/json',
             ],
@@ -87,5 +87,7 @@ class Provider extends AbstractProvider
                 'refresh_token' => $refreshToken
             ]
         ]);
+        
+        return json_decode((string) $response->getBody(), true);
     }
 }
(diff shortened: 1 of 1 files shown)


src/HubSpot/Provider.php
```diff
@@ -63,4 +63,29 @@ protected function mapUserToObject(array $user)
...
+     * @see https://developers.hubspot.com/docs/api/oauth-quickstart-guide#refreshing_tokens
+     *
+     * @param string $refreshToken
+     * @return ResponseInterface
+     *
+     * @throws \GuzzleHttp\Exception\GuzzleException
+     */
+    public function refreshToken($refreshToken): ResponseInterface
+    {
+        return $this->getHttpClient()->post($this->getTokenUrl(), [
```
lucasmichot, line 79:
```suggestion
        $response = $this->getHttpClient()->post($this->getTokenUrl(), [
```

src/HubSpot/Provider.php
```diff
@@ -63,4 +63,29 @@ protected function mapUserToObject(array $user)
...
         ]);
     }
+
+    /**
+     * Acquire a new access token using the refresh token.
+     *
+     * @see https://developers.hubspot.com/docs/api/oauth-quickstart-guide#refreshing_tokens
+     *
+     * @param string $refreshToken
+     * @return ResponseInterface
```
lucasmichot, line 73:
```suggestion
     * @return array
```

src/HubSpot/Provider.php
```diff
@@ -63,4 +63,29 @@ protected function mapUserToObject(array $user)
...
+            RequestOptions::HEADERS => [
+                'Accept' => 'application/json',
+            ],
+            RequestOptions::FORM_PARAMS => [
+                'client_id' => $this->clientId,
+                'client_secret' => $this->clientSecret,
+                'grant_type' => 'refresh_token',
+                'refresh_token' => $refreshToken
+            ]
+        ]);
```
lucasmichot, line 89:
```suggestion
        ]);
        
        return json_decode((string) $response->getBody(), true);
```

lucasmichot (changes requested):
Can you return an array instead

dansleboby:
Yes, but we should change Apple too to keep a standard

atymic (approved):

Issue #1165, opened 2024-03-04
AzureADB2C has error in emails mapping


baradhili:
currently this is in Provider
```
protected function mapUserToObject(array $user)
    {
        return(new User())->setRaw($user)->map([
            'id' => $user['sub'],
            'nickname' => $user['name'],
            'name' => $user['name'],
            'email' => $user['emails'][0],
        ]);
    }
```
However Azure OpenID returns:
```
[ver] => 1.0
    [iss] => 
    [sub] => 
    [aud] => 
    [exp] => 
    [acr] => b2c_1a_dev_ob_shared_corpaad_basic_signin
    [iat] => 
    [auth_time] => 
    [socialIdpUserId] => 
    [given_name] => 
    [family_name] => 
    [email] => 
    [name] => 
    [authenticationSource] => 
    [idp] => 
    [nbf] => 
```

code should be 
```
protected function mapUserToObject(array $user)
    {
        return(new User())->setRaw($user)->map([
            'id' => $user['sub'],
            'nickname' => $user['name'],
            'name' => $user['name'],
            'email' => $user['email'],
        ]);
    }
```

Though I suspect this might also be a MS barf and I wouldn't be surprised if they return different claims depending on the policy type etc.. might need to do some decisions before mapping

baradhili:
note socialIdpUserId also contains the email

atymic:
Can you link to the docs please? PR welcome

baradhili:
PR sent

PR #1171, opened 2024-03-08
Update Azure ADB2C for OpenId differences

### LINKED ISSUE #1320: Microsoft Socialite provider 4.6.0 contains breaking changes
In version 4.6.0 explicit typehints were introduced in these commits:
- https://github.com/SocialiteProviders/Microsoft/commit/76d51308ef53d1425a5ce3c3b094b0925cd82951
- https://github.com/SocialiteProviders/Microsoft/commit/87e61284d17241f5b65e28797239839ca1eaf5c3

Since the typehints are mandatory in all child classes these changes should be reverted in favor of docblocks and introduced only in v5.0.

baradhili:
This seems to work for me. Have included URL I relied on for scope space separator as well as handling if the $user['email'] is an array or not since there is no spec from MS saying it will always be a scalar
fixes #1320

### DIFF
--- a/src/AzureADB2C/Provider.php
+++ b/src/AzureADB2C/Provider.php
@@ -23,6 +23,12 @@ class Provider extends AbstractProvider
         'openid',
     ];
 
+    /**
+     * The scope separator used for Azure AD B2C authentication.
+     * https://learn.microsoft.com/en-us/azure/active-directory-b2c/openid-connect#get-a-token:~:text=example%20b2c_1_sign_in.-,HTTP,-Copy
+     * 
+     * @var string
+     */
     protected $scopeSeparator = ' ';
 
     /**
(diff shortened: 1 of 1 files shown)


codespearhead:
Closes #1165

src/AzureADB2C/Provider.php
```diff
@@ -198,11 +204,18 @@ private function validateIdToken($idToken)
      */
     protected function mapUserToObject(array $user)
     {
+        //check is $user['email'] is array - it shouldn't be, but we are not sure
+        if (is_array($user['email'])) {
```
atymic, line 208: I guess we should also make sure this is not empty? Or alt, just do
`$user['email'][0] ?? $user['email']`

atymic: No worries.

baradhili:
Looks like master has updated to handle this.

PR #1172, opened 2024-03-11
Clover regions

macbookandrew:
Adds support for Europe and Latin America regions

### DIFF
--- a/src/Clover/README.md
+++ b/src/Clover/README.md
@@ -52,7 +52,21 @@ The user includes a `token` property that you can use to retrieve the API token 
 Route::get('clover/auth/callback', function () {
     $user = Socialite::driver('clover')->user();
 
-    // Save this token somewhere for other use.
-    $token = $user->token;
+    // Save these tokens somewhere for use with the API.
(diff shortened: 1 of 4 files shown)


src/Clover/README.md
```diff
@@ -17,7 +17,7 @@ Ensure the app has permission to read employees.
   'client_id' => env('CLOVER_CLIENT_ID'),
   'client_secret' => env('CLOVER_CLIENT_SECRET'),
   'redirect' => env('CLOVER_REDIRECT_URI')
-  'environment' => env('CLOVER_ENVIRONMENT', 'production'),
+  'environment' => env('CLOVER_ENVIRONMENT', 'production'), // one of the following: 'sandbox', 'production' (for US/Canada), 'europe', or 'latin-america'
```
atymic, line 20: Can we put `united_states` instead to keep it consistent?

macbookandrew: Sure, but I used `north_america` instead as it covers US and Canada

atymic (approved):

PR #1177, opened 2024-03-20
fix: instagram provider merge code fields

kife-design:
Align the getCodeFields method with the AbstractProvider.php where the CodeFields are merged with the external parameters.

### DIFF
--- a/src/InstagramBasic/Provider.php
+++ b/src/InstagramBasic/Provider.php
@@ -86,15 +86,11 @@ class Provider extends AbstractProvider
      */
     protected function getCodeFields($state = null)
     {
-        $fields = [
-            'state'         => $state,
-            'response_type' => 'code',
-            'app_id'        => $this->clientId,
-            'redirect_uri'  => $this->redirectUrl,
-            'scope'         => $this->formatScopes($this->scopes, $this->scopeSeparator),
-        ];
-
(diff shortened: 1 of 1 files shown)


kife-design:
We are using instagrambasic and instagram graph api in one project. They share the same code, and a custom parameter is passed to the redirect using the with() method.

In the instagram basic provider, the custom param is not passed because it doe not merge the code fields with the provided parameters. This pull request fixes that issue.

src/InstagramBasic/Provider.php
```diff
@@ -86,13 +86,15 @@ protected function mapUserToObject(array $user)
      */
     protected function getCodeFields($state = null)
```
lucasmichot, line 87: I guess this whole method should just be:
```php
protected function getCodeFields($state = null): array
{
    $fields = parent::getCodeFields($state);
    $fields['app_id'] = $fields['client_id'];
    unset($fields['client_id']);
    
    return $fields;
}
```

lucasmichot (changes requested):

src/InstagramBasic/Provider.php
kife-design: Updated

lucasmichot (approved):

atymic (approved):

PR #1179, opened 2024-03-26
Update LifeScienceLogin config URL

mzur:
LSLogin v2 uses a new config URL now. Services can be self-managed via  https://services.aai.lifescience-ri.eu. Once you upgrade to the new provider version, you have to update the `LSLOGIN_CLIENT_ID` and `LSLOGIN_CLIENT_SECRET`, too.


lucasmichot (approved):

atymic:
@mzur does this need a major version tag if you need to update credentials?

mzur:
Yes, this can be considered a breaking change.

PR #1180, opened 2024-03-26
Update Azure ADB2C user email field

vincent-paqt:
Azure ADB2C can return a `email` key. This PR checks if there is a `emails` array or a singel `email` key.

### DIFF
--- a/src/AzureADB2C/Provider.php
+++ b/src/AzureADB2C/Provider.php
@@ -202,7 +202,7 @@ class Provider extends AbstractProvider
             'id'       => $user['sub'],
             'nickname' => $user['name'],
             'name'     => $user['name'],
-            'email'    => $user['emails'][0],
+            'email'    => $user['emails'][0] ?? $user['email'],
         ]);
     }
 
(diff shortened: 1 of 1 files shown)


src/AzureADB2C/Provider.php
```diff
@@ -202,7 +202,7 @@ protected function mapUserToObject(array $user)
             'id'       => $user['sub'],
             'nickname' => $user['name'],
             'name'     => $user['name'],
-            'email'    => $user['emails'][0],
+            'email'    => $user['emails'][0] ?? $user['email'],
```
lucasmichot, line 205: If none of them are set, shall this fall back to `null` ?

vincent-paqt: @lucasmichot Yeah, check out: https://3v4l.org/p3Fb4

```php
<?php

$user = [];

var_dump($user['emails'][0] ?? $user['email']); //returns NULL
```

atymic: But this emits a warning. Lets do `$user['emails'][0] ?? $user['email'] ?? null`?

atymic (changes requested):

atymic (approved):

PR #1183, opened 2024-03-28
Make TikTok username optional based on scopes set

Skullbock:
In v2, `username` can be requested only if `user.info.profile` scope is set.
Instead of forcing the scope, since it's not even in the default list of scopes set in the provider, we could just add the username field if the scope is present.

Closes #1198

PR #1199, opened 2024-04-25
Update TikTok Provider requested scopes

nexxai:
As per https://laracasts.com/discuss/channels/laravel/issue-with-socialite-tiktok we also need access to the `user.info.profile` scope


atymic:
Closed in favor of #1183

PR #1201, opened 2024-05-07
docs: update linkdin provider documentation

TilsonM17:
Recently I had an error when trying to do Sig In, the reason is that Linkdin changed the way it authenticates, and the package is already that way.

The old way to connect with provider givem error, because now LinkedIn is using Sign In with LinkedIn using OpenID Connect instead of Sign In with LinkedIn.

So you just need to use linkedin-openid instead of linkedin in the `service.php` and `Socialite::driver('linkedin-openid')->redirect();`

### DIFF
--- a/src/LinkedIn/README.md
+++ b/src/LinkedIn/README.md
@@ -11,7 +11,7 @@ Please see the [Base Installation Guide](https://socialiteproviders.com/usage/),
 ### Add configuration to `config/services.php`
 
 ```php
-'linkedin' => [    
+'linkedin-openid' => [   
   'client_id' => env('LINKEDIN_CLIENT_ID'),  
   'client_secret' => env('LINKEDIN_CLIENT_SECRET'),  
   'redirect' => env('LINKEDIN_REDIRECT_URI') 
@@ -54,7 +54,7 @@ protected $listen = [
 You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
 
 ```php
-return Socialite::driver('linkedin')->redirect();
+return Socialite::driver('linkedin-openid')->redirect();
 ```
 
 ### Returned User fields


atymic:
Are you sure this is correct?
https://github.com/SocialiteProviders/Providers/blob/master/src/LinkedIn/LinkedInExtendSocialite.php

atymic:
Closing due to inactivity

quantumwebco:
> Are you sure this is correct? https://github.com/SocialiteProviders/Providers/blob/master/src/LinkedIn/LinkedInExtendSocialite.php

I can confirm this is correct. Been struggling getting linkedin to work due to the package using the deprecated API and scopes. Using 'linkedin-openid' as the provider name and config fixed it and now it's working as expected

quantumwebco:
Digging around more in relation to this, it is really confusing with the issues and PRs. I think it should state in the docs that using 'linkedin' will not work, maybe it works for legacy, but as far as I know they have deprecated it completely now, so you have to use 'linkedin-openid'

atymic:
Could you PR an update to the docs? And then i can deprecate the old one.

quantumwebco:
This PR literally does that but you closed it

atymic:
It's not our provider, is what I mean. you're using the base socialite provider if you use that
https://github.com/laravel/socialite/blob/1e177bf35a644db34c207b44d752c0d8e8d7c004/src/Two/LinkedInOpenIdProvider.php

TilsonM17:
@atymic  

The Linkedin current use v2 for they api, and only accept openID con sign in in API.

Try to create a new project and try sign-in with  current instructions in docs. You will get an error.

Socialite package are correct they're implementing with new way, the right way . The only problem is docs. Docs need update. And this PR does exactly this.


https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin-v2

quantumwebco:
The Linkedin driver in this package is deprecated then, completely defunct, and should be removed from the docs entirely, right? Or is it better to leave the page there but refer back to the core documentation of using 'linkedin-openid', as per https://laravel.com/docs/11.x/socialite#configuration?

PR #1214, opened 2024-06-09
feat: zettle provider

audunru:
Zettle by Paypal is a point of sale payment provider.

### DIFF
--- a/src/Zettle/README.md
+++ b/src/Zettle/README.md
@@ -1,0 +1,66 @@
+# Zettle
+
+Zettle by Paypal is a point of sale payment provider. [API documentation](https://developer.zettle.com/docs/api)
+
+```bash
+composer require socialiteproviders/zettle
+```
+
+## Installation & Basic Usage
+
+Please see the [Base Installation Guide](https://socialiteproviders.com/usage/), then follow the provider specific instructions below.
+
+### Add configuration to `config/services.php`
+
+```php
+'zettle' => [
+  'client_id' => env('ZETTLE_CLIENT_ID'),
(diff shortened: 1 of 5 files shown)


src/Zettle/README.md
```diff
@@ -0,0 +1,66 @@
...
+        // ... other providers
+        \SocialiteProviders\Zettle\ZettleExtendSocialite::class.'@handle',
+    ],
+];
+```
+</details>
+
+### Usage
+
+You should now be able to use the provider like you would regularly use Socialite (assuming you have the facade installed):
```
atymic, line 56: Can you add a comment to mention that the provider only returns a uuid id?

src/Zettle/composer.json
```diff
@@ -0,0 +1,33 @@
...
+            "email": "audun@rundberg.no"
+        }
+    ],
+    "support": {
+        "issues": "https://github.com/socialiteproviders/providers/issues",
+        "source": "https://github.com/socialiteproviders/providers",
+        "docs": "https://socialiteproviders.com/zettle"
+    },
+    "require": {
+        "php": "^8.0",
```
atymic, line 24:
```suggestion
        "php": "^8.1",
```

src/Zettle/README.md
audunru: I added a section about it. [According to the docs](https://developer.zettle.com/docs/api/oauth/api-reference#responses-7) you only get a user ID and organization ID. There is no scope to get more details AFAIK.

atymic (approved):

PR #1217, opened 2024-06-21
Add Threads provider

nickbeen:
This PR adds a provider for Threads (www.threads.net), the Twitter-like social network of Meta.

Note: Threads does not support refresh tokens, but allows the default short-lived access tokens to be exchanged for long-lived refreshable access tokens. I've added example code in the README.md instead of incorrectly implementing it in the `getRefreshTokenResponse` method.

Docs: https://developers.facebook.com/docs/threads


atymic (approved):

Pr4w:
Hi,

Found a small issue with this. If a user doesn't have a profile picture then the callback breaks, since the key "threads_profile_picture_url" doesn't exist on the $user array. I fixed it locally, but you might want to check if it exists and revert to null if it doesn't.

See line 84 of Provider.php.

atymic:
https://github.com/SocialiteProviders/Providers/pull/1230

PR #1220, opened 2024-06-25
Add userinfo_uri configurable. Fixed #1219

### LINKED ISSUE #1219: can we make the userinfo_uri configurable?
Hi, I have a requirement to deploy Laravel to Vercel, which used /api , so the Laravel API
 I changed to v1/api/* , so I need change the LaravelPassport userinfo_uri to 'v1/api/user'.
so can we make the userinfo_uri configurable? thanks. 

https://github.com/SocialiteProviders/Providers/blob/1cf2b81f84a85e89f8c96ed1ad7344d1fd3ae3f4/src/LaravelPassport/Provider.php#L107

guoxiangke:
This pull make userinfo_uri configurable in socialiteproviders/laravelpassport.

### DIFF
--- a/src/LaravelPassport/Provider.php
+++ b/src/LaravelPassport/Provider.php
@@ -104,7 +104,7 @@ class Provider extends AbstractProvider
         return rtrim($this->getConfig('host'), '/').'/'.ltrim($this->getConfig($type, Arr::get([
             'authorize_uri' => 'oauth/authorize',
             'token_uri'     => 'oauth/token',
-            'userinfo_uri'  => 'api/user',
+            'userinfo_uri'  => $this->getConfig('userinfo_uri')',
         ], $type)), '/');
     }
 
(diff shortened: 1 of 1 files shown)


src/LaravelPassport/Provider.php
```diff
@@ -104,7 +104,7 @@ protected function getLaravelPassportUrl($type)
         return rtrim($this->getConfig('host'), '/').'/'.ltrim($this->getConfig($type, Arr::get([
             'authorize_uri' => 'oauth/authorize',
             'token_uri'     => 'oauth/token',
-            'userinfo_uri'  => 'api/user',
+            'userinfo_uri'  => $this->getConfig('userinfo_uri')',
```
atymic, line 107: Can you add the default value? otherwise it will be null if you don't set that config.

atymic (changes requested):

PR #1221, opened 2024-06-26
Fixing APS OAuth

SuperDJ:
Fixing APS OAuth.

Also added possibility to use `setConfig`. This makes it possible to use the package in a multi-tenancy app. 

According to the [docs](https://aps.autodesk.com/en/docs/oauth/v2/tutorials/get-3-legged-token/#step-3-exchange-the-authorization-code-for-an-access-token) a base64 encrypted header of the client ID and secret is required. But not all "standard" headers are required or matched the docs. For example the content-type.

### DIFF
--- a/src/AutodeskAPS/Provider.php
+++ b/src/AutodeskAPS/Provider.php
@@ -35,6 +35,61 @@ class Provider extends AbstractProvider
         return 'https://developer.api.autodesk.com/authentication/v2/token';
     }
 
+	/**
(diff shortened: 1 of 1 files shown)


atymic:
@SuperDJ can we get a little more explaination here?

SuperDJ:
@atymic updated the description. Let me know if you need more information.

atymic (approved):

PR #1225, opened 2024-06-28
feat: add Aikido OAuth support

olivernybroe:
This PR adds support for Aikido's OAuth provider
https://apidocs.aikido.dev/ 

Aikido does not have a user endpoint, but have a workspace endpoint instead. An authentication is tied to the workspace, not to the specific user. 

Aikido is also one of the providers which uses `Basic` auth in the headers for client and secret for their token requests.


lucasmichot (approved):
GTM @atymic

atymic (approved):

olivernybroe:
@lucasmichot The provider is not working anymore as the `Basic` auth header was removed from my original pr.

olivernybroe:
Created a new pr to fix it #1227

PR #1227, opened 2024-07-01
fix: Aikido - add basic auth header for token request

olivernybroe:
Adding back the Basic auth header for the token request which was removed as part of the reviewing process. 

Without it the following error will happen
```
    Client error: `POST https://app.aikido.dev/api/oauth/token` resulted in a `401 Unauthorized` response:
    {"error":"invalid_client","error_description":"The request is missing an Authorization header"}
```

### DIFF
--- a/src/Aikido/Provider.php
+++ b/src/Aikido/Provider.php
@@ -41,6 +41,16 @@ class Provider extends AbstractProvider
     }
 
     /**
+     * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


olivernybroe:
ping @lucasmichot

lucasmichot (approved):

atymic (approved):

PR #1232, opened 2024-07-08
Fixes #1231; Null tenant info being returned. 

### LINKED ISSUE #1231: Tenant Returning Null On Microsoft Provider Even With include_tenant_info set to true.
Hi there,

I recently encountered an issue while using the Microsoft provider where the tenant is being returned as `null` even though I have ` include_tenant_info` set to true on my configuration.

Here is what I have in my services.php:
```php
    'microsoft' => [
        'client_id' => env('MICROSOFT_CLIENT_ID'),
        'client_secret' => env('MICROSOFT_CLIENT_SECRET'),
        'redirect' => env('MICROSOFT_REDIRECT_URI'),
        'include_tenant_info' => true,
        'tenant' => "organizations"
    ],
```
I have tried two separate Microsoft enterprise accounts on completely separate tenants, and both times they returned null. In addition I have tried several different applications (different client IDs/secrets) and the issue still occurs.

Is there an extra permission I need to enable to get this to work? I had it working a couple of weeks ago and I did not make any changes so I'm curious if MS maybe updated their API, or perhaps something needs to change on the organization side in terms of permissions. 

Let me know if any more context is needed. Thank you!

jortuck:
I encountered an issue when using `include_tenant_info` where the tenant info being returned was null every time. The issue was caused by a forced use of the common tenant when fetching tenant info.
```php
$this->getConfig('tenant', 'common') === 'common'
```
I can't find a reason for why this was here, as it prevents users who are using a custom tenant ID or the `organizations` tenant from fetching tenant info. I can confirm tenant info began being returned again after removing the check. It still works with the `common` tenant, `organizations`  tenant, `consumers` tenant, and a custom `tenant identifier`.

In addition, when a user logs in with an account where the tenant does not exists (ex a personal account), Microsoft returns a 500 error. Currently, the exception being caught is a `ClientException` when it should be a `ServerException`. This fixes the 500 error when users login with a consumer account.

This will close #1231

### DIFF
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -3,6 +3,7 @@
 namespace SocialiteProviders\Microsoft;
 
 use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\ServerException;
 use GuzzleHttp\RequestOptions;
 use Illuminate\Support\Arr;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
@@ -149,7 +150,7 @@ class Provider extends AbstractProvider
                     ]
                 );
                 $formattedResponse['tenant'] = json_decode((string) $responseTenant->getBody(), true)['value'][0] ?? null;
-            } catch (ClientException) {
+            } catch (ServerException) {
                 //if exception then tenant does not exist.
                 $formattedResponse['tenant'] = null;
             }
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -3,7 +3,7 @@
 namespace SocialiteProviders\Microsoft;
 
 use GuzzleHttp\Exception\ClientException;
-use GuzzleHttp\Exception\ServerException;
+use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\RequestOptions;
 use Illuminate\Support\Arr;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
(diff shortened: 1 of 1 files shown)


src/Microsoft/Provider.php
```diff
@@ -149,7 +150,7 @@ protected function getUserByToken($token)
                     ]
                 );
                 $formattedResponse['tenant'] = json_decode((string) $responseTenant->getBody(), true)['value'][0] ?? null;
-            } catch (ClientException) {
+            } catch (ServerException) {
```
atymic, line 153: Should we not catch both here? Or will it only ever return a 500 in the case the tenant does not exist?

jortuck: I am not entirely sure because Microsoft doesn't have great documentation on what specific HTTP error codes are produced and when. From my experience, it as only ever returned 500 when the tenant does not exists. It could be related to this PR [https://github.com/SocialiteProviders/Providers/pull/1035](https://github.com/SocialiteProviders/Providers/pull/1035/)

I actually like your idea of catching both exceptions. My theory is that the `ClientException` occurs when the the token you use the fetch the tenant is not allowed to access information from that tenant (caused by trying to get a tenant using a token from a user who is not a member of said tenant). The `ServerException` happens when the user has no tenant (ex. a consumer account) or the tenant does not exists at all. Either way I can't find any documentation about it so I have no way to confirm without testing.

atymic: Maybe just catch `RequestException` which will cover both cases?

jortuck: Sounds good, I will go ahead and add that now!

atymic (approved):

PR #1245, opened 2024-08-18
Add Zitadel provider

tbhaxor:


src/Zitadel/composer.json
```diff
@@ -0,0 +1,28 @@
...
+        }
+    },
+    "authors": [
+        {
+            "name": "Gurkirat Singh",
+            "email": "tbhaxor@gmail.com"
+        }
+    ],
+    "require": {
+        "php": "^8.0",
```
atymic, line 19:
```suggestion
        "php": "^8.1",
```

tools/docgen.php
```diff
@@ -62,10 +62,11 @@
 
 DOC;
 
-$directories = array_map('basename', glob('../src'.'/*', GLOB_ONLYDIR));
+$providersDir = realpath(__DIR__ . '/../src');
+$directories = array_map('basename', glob($providersDir . '/*', GLOB_ONLYDIR));
 
 foreach ($directories as $provider) {
-    $path = sprintf('%s/../src/%s/README.md', __DIR__, $provider);
```
atymic, line 68: Was there a reason for this change?

tbhaxor: Yes, which this you can run the docgen file from any directory. For example,

```console
php ../../tools/docgen.php
```

tbhaxor:
@atymic The provider is not available on packagist.com and also the readme contents are not available on the website https://socialiteproviders.com/Zitadel/.

atymic:
You'll need to PR to the website for the second one, see email for first.

PR #1247, opened 2024-08-20
fix: threads profile picture url if not available and update readme

Akmalfauzi:


atymic:
I don't see any changes here. Arr::get already returns null?

Akmalfauzi:
Yes, that's right, because in my case, it only returns the id and username fields.

So, I fixed the undefined index "threads_profile_picture_url" error so that it could still run even though the "threads_profile_picture_url" field did not exist or was empty.

Akmalfauzi:
I also fixed the typo in the `README.md` file when adding code to the `EventServiceProvider.php` file

Screenshot 2024-08-21 at 08 06 42

atymic:
`Arr::get` will return null already if the field does not exist?

Akmalfauzi:
Yes, that's right.

src/Threads/Provider.php
```diff
@@ -82,7 +82,7 @@ protected function mapUserToObject(array $user)
             'nickname'     => Arr::get($user, 'username'),
             'name'         => null,
             'email'        => null,
-            'avatar'       => Arr::get($user, 'threads_profile_picture_url'),
+            'avatar'       => Arr::get($user, 'threads_profile_picture_url') ?? null,
```
atymic, line 85:
```suggestion
            'avatar'       => Arr::get($user, 'threads_profile_picture_url'),
```

No need to add the null coalesce

atymic (changes requested):
Just this change then it's good merge.

src/Threads/Provider.php
Akmalfauzi: hmm, oke thank you. i will fix soon.

atymic (approved):

PR #1255, opened 2024-08-26
Lint all files

lucasmichot:
Run `pint` to lint all files


atymic:
Have run the action :)

PR #1256, opened 2024-08-27
Add extra pint rule

lucasmichot:
Add ab extra `pint` rule


atymic (commented):
what's the reasoning for this change?

atymic:
IMO It's not worth polluting the changes history of all the files for this change

PR #1259, opened 2024-09-03
Microsoft - Update email field to return mail

JosefHarney:
The base package `laravel/socialite` has the method `getEmail` within the class [AbstractUser](https://github.com/laravel/socialite/blob/5.x/src/AbstractUser.php#L94). Within `mapUserToObject` it is returning `userPrincipalName` in the email field, therefor `getEmail` is returning an empty string.

This PR is to return the correct field. 

Side note the `userPrincipalName` is return on #195 of the provider.


atymic (commented):
Can you explain the reason for this change? It would be a breaking change, but also it is my understanding the `mail` property can be changed by the user.

Issue #1115, opened 2023-11-12
Linkedin: Add Linkedin OpenId Socialite Provider.


akhterarif:
As Linkedin has introduced openid for auth, Socialite also made a new provider for `linkedin-openid`. It is not available here. It needs to be added here also.

atymic:
Is there a reason we can't just use the upstream one?

PR #1263, opened 2024-09-14
Use LinkedIn sign in with Open ID 

rcknr:
Since August 2023 LinkedIn [deprecated](https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin) their older sign-in which is still used here. The new provider was added to Socialite so it should be used here.
#1115


Baspa:
Can this be merged please? The socialite package is not working for LinkedIn now @lucasmichot @AndyWendt @m1guelpf @atymic

PR #1275, opened 2024-10-22
Add the children requirements to the root composer.json

lucasmichot:
Add the children requirements to the root composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,12 @@
     "require": {
         "php": "^8.0",
         "ext-json": "*",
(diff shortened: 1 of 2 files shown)


atymic:
What is the reasoning for this?

lucasmichot:
We need to ensure that we can still install all dependencies, depending on the PHP version

atymic (commented):
If we're going to do this, we need to update the contribution guidelines to make sure people add it in their PRs. In addition, we might end up with conflicts between different drivers, how would we resolve that?

PR #1277, opened 2024-10-25
Add Nuvemshop Custom Provider

jmmidia:
Very similar to existent MercadoLibre but it changes the auth URL.

Docs: https://tiendanube.github.io/api-documentation/authentication

### DIFF
--- a/src/Nuvemshop/Provider.php
+++ b/src/Nuvemshop/Provider.php
@@ -83,10 +83,8 @@ class Provider extends AbstractProvider
     {
         return (new User)->setRaw($user)->map([
             'id' => $user['user_id'],
-            'nickname' => null,
-            'name' => null,
-            'email' => $user['email'],
-            'avatar' => null
+            'name' => $user['name'],
+            'email' => $user['email']
         ]);
     }
 }
--- a/src/Nuvemshop/Provider.php
+++ b/src/Nuvemshop/Provider.php
@@ -61,13 +61,14 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
-    protected function getUserByToken($token)
+    protected function getUserByToken($token, $userAgent)
     {
         $response = $this->getHttpClient()->get(
             'https://api.nuvemshop.com.br/v1/store',
             [
                 RequestOptions::HEADERS => [
-                    'Authentication' => 'Bearer ' . $token
+                    'Authentication' => 'Bearer ' . $token,
+                    'User-Agent' => $userAgent
                 ]
             ]
         );
--- a/src/Nuvemshop/Provider.php
+++ b/src/Nuvemshop/Provider.php
@@ -61,13 +61,16 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
-    protected function getUserByToken($token, $userId)
+    protected function getUserByToken($token)
     {
+        $accessToken = explode('.', $token)[0];
+        $userId = explode('.', $token)[1];
+
         $response = $this->getHttpClient()->get(
             "https://api.nuvemshop.com.br/v1/$userId/store",
             [
                 RequestOptions::HEADERS => [
-                    'Authentication' => 'bearer ' . $token,
+                    'Authentication' => 'bearer ' . $accessToken,
                     'User-Agent' => $this->getClientId()
                 ]
             ]
@@ -84,7 +87,9 @@ class Provider extends AbstractProvider
         return (new User)->setRaw($user)->map([
             'id' => $user['id'],
             'name' => $user['name'],
-            'email' => $user['email']
+            'nickname' => null,
+            'email' => $user['email'],
+            'avatar' => null
         ]);
     }
 }
--- a/src/Nuvemshop/Provider.php
+++ b/src/Nuvemshop/Provider.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace SocialiteProviders\Nuvemshop;
+
+use GuzzleHttp\RequestOptions;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    public const IDENTIFIER = 'NUVEMSHOP';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state)
+    {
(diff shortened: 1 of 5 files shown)


gabrielrbarbosa (approved):
lgtm, @atymic can you review this PR please?

Nuvemshop info here:
https://techcrunch.com/2020/10/13/nuvemshop-a-latin-american-answer-to-shopify-raises-30-million/

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,108 @@
...
+        'read_orders',
+        'write_orders',
+        'read_coupons',
+        'write_coupons',
+        'write_scripts',
+        'write_shipping'
+    ];
+
+    /**
+     * The separating character for the requested scopes.
```
lucasmichot, line 36: Please use `{@inheritdoc}` here

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,108 @@
...
+     * @var string
+     */
+    protected $scopeSeparator = ' ';
+
+    /**
+     * Returns the configured user id that we're authenticating with
+     *
+     * @return string
+     */
+    private function getClientId()
```
lucasmichot, line 47: This is useless: the `clientId` is already available as class property. Please remove it and update the code accordingly.

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return 'https://www.tiendanube.com/apps/authorize/token';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getUserByToken($token)
+    {
+        $accessToken = explode('.', $token)[0];
```
lucasmichot, line 79: Just use `[$bearerToken, $userId, $_] = explode('.', $token, 3);`

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,108 @@
...
+        $userId = explode('.', $token)[1];
+
+        $response = $this->getHttpClient()->get(
+            "https://api.nuvemshop.com.br/v1/$userId/store",
+            [
+                RequestOptions::HEADERS => [
+                    'Authentication' => 'bearer ' . $accessToken,
```
lucasmichot, line 86:
```suggestion
                    'Authentication' => 'Bearer ' . $accessToken,
```

lucasmichot (changes requested):

src/Nuvemshop/Provider.php
jmmidia: It should be lowercase as the [documentation](https://tiendanube.github.io/api-documentation/authentication) suggests: "The entire bearer word should be lowercase. Using a different word will result in a 401: unauthorized - invalid access token error."

lucasmichot: Up to you. Nevertheless, I would probably expect any server to support both lowercase and propercase. As lowercase is pretty uncommon.
cf. `\Illuminate\Http\Concerns\InteractsWithInput::bearerToken`

jmmidia: I made two requests using Insomnia, one with lowercase and another with uppercase and received the following:

Uppercase:
image

Lowercase:
image

src/Nuvemshop/Provider.php
jmmidia: Shouldn't it be [$bearerToken, $userId] = explode('.', $token, 2)? (it worked)

I tested it in my local environment and it returned me an error: "Undefined Array Key 2". Since I'm expecting a single string with both access_token and user_id separated by a dot, the limit parameter should be 2.

lucasmichot: I had the feeling that `$token` had the shape of a JWT.
Of couse, if the token contains only one separator dot, `explode('.', $token, 2)` is fine.

lucasmichot:
@jmmidia make sure to also create a PR for https://github.com/SocialiteProviders/website

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace SocialiteProviders\Nuvemshop;
+
+use GuzzleHttp\Exception\GuzzleException;
```
lucasmichot, line 5: Unused, please remove

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,95 @@
...
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state)
+    {
+        return $this->buildAuthUrlFromBase(
```
lucasmichot, line 45: `return $this->buildAuthUrlFromBase("https://www.tiendanube.com/apps/{$this->clientId}/authorize", $state);`

is just fine

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state)
```
lucasmichot, line 43:
```suggestion
    protected function getAuthUrl($state): string
```

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,95 @@
...
+                $this->clientId
+            ),
+            $state
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getTokenUrl()
```
lucasmichot, line 57:
```suggestion
    protected function getTokenUrl(): string
```

src/Nuvemshop/Provider.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        return 'https://www.tiendanube.com/apps/authorize/token';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getUserByToken($token)
```
lucasmichot, line 65:
```suggestion
    protected function getUserByToken($token): array
```

lucasmichot (commented):
Some more changes

lucasmichot (approved):
GTM @atymic

atymic (approved):

PR #1278, opened 2024-10-29
package name fix

amdad121:
It was `socialiteproviders/paypalsandbox` but original package is `socialiteproviders/paypal-sandbox`. so i changed this.


lucasmichot (approved):
cf. https://packagist.org/packages/socialiteproviders/paypal-sandbox
GTM @atymic

atymic (approved):

PR #1282, opened 2024-10-29
Normalize all `additionalConfigKeys` methods

lucasmichot:
Normalize all `additionalConfigKeys` methods

### DIFF
--- a/src/AzureADB2C/Provider.php
+++ b/src/AzureADB2C/Provider.php
@@ -218,9 +218,6 @@ class Provider extends AbstractProvider
             .urlencode($post_logout_uri);
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public static function additionalConfigKeys(): array
     {
         return [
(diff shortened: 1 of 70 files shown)


src/AzureADB2C/Provider.php
```diff
@@ -219,9 +219,9 @@ public function logout($post_logout_uri)
     }
 
     /**
-     * @return array
+     * {@inheritdoc}
```
lucasmichot, line 222: @atymic cf. https://github.com/SocialiteProviders/Providers/pull/1284#discussion_r1821659496
Maybe we can remove the docblocks too, let me know

atymic: Yep may as well if you're changing it anyway

lucasmichot: I've removed them; see 09a0576908210bd9fad5eb7f762555cf88f6293d

atymic (approved):

PR #1284, opened 2024-10-29
Normalize all `scope` and `scopeSeparator` properties

lucasmichot:
Normalize all `scope` and `scopeSeparator` properties

### DIFF
--- a/src/Line/Provider.php
+++ b/src/Line/Provider.php
@@ -12,16 +12,12 @@ class Provider extends AbstractProvider
     public const IDENTIFIER = 'LINE';
 
     /**
-     * The separating character for the requested scopes.
-     *
-     * @var string
+     * {@inheritdoc}
      */
     protected $scopeSeparator = ' ';
 
     /**
-     * The scopes being requested.
-     *
-     * @var array
+     * {@inheritdoc}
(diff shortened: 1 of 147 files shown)


src/Line/Provider.php
```diff
@@ -12,16 +12,12 @@ class Provider extends AbstractProvider
...
-     * @var string
+     * {@inheritdoc}
      */
     protected $scopeSeparator = ' ';
 
     /**
-     * The scopes being requested.
-     *
-     * @var array
+     * {@inheritdoc}
```
atymic, line 20: Why is this needed? This is just for code doc generation right, which we dont do?

lucasmichot: > This is just for code doc generation right, which we dont do?
It is

Another approach would be be to remove all the docblocks

lucasmichot: @atymic I can remove them all, just let me know

lucasmichot: I've removed them; see 3de8aca127600ef506df292e3ca66d06a62cb202

atymic (approved):

atymic (approved):

PR #1285, opened 2024-10-29
Use `RequestOptions::QUERY` when possible

lucasmichot:
Use `RequestOptions::QUERY` when possible

### DIFF
--- a/src/TeamService/Provider.php
+++ b/src/TeamService/Provider.php
@@ -35,14 +35,16 @@ class Provider extends AbstractProvider
      */
     protected function getUserByToken($token)
     {
-        $response = $this->getHttpClient()->get(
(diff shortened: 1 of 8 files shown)


src/TeamService/Provider.php
```diff
@@ -35,14 +35,16 @@ protected function getTokenUrl()
...
-            ]
-        );
+        $uri = 'https://'.$this->getAccount().'.vssps.visualstudio.com/_apis/profile/profiles/me';
+
+        $response = $this->getHttpClient()->get($uri, [
+            RequestOptions::HEADERS => [
+                'Authorization' => 'Bearer '.$token,
+            ],
+            RequestOptions::QUERY => [
+                'api-version' => '5.0-preview.3',
```
atymic, line 45: Do we need to consider any differences in URL encoding here?

lucasmichot: No

PR #1299, opened 2024-11-24
Fix Threads - no avatar default value

adam-riha:
Fix for account connection if no avatar is set.

Property avatar cannot be null.


atymic:
What do you mean?
Please provide the code, `Arr::get` will just return null

adam-riha:
OK my bad, sorry.

I overlooked that there is no defined type, 
but only string phpdoc in `Laravel\Socialite\AbstractUser` for `$avatar` property.

PR #1304, opened 2024-12-02
Change the function visibility of Zendesk's getSubdomain

27pchrisl:
Hi, I'm writing an integration for Zendesk that works globally, so I need to be able to change the subdomain dynamically. I extended the Provider but need to override this method 😄

### DIFF
--- a/src/Zendesk/Provider.php
+++ b/src/Zendesk/Provider.php
@@ -60,7 +60,7 @@ class Provider extends AbstractProvider
      *
      * @return string
      */
-    private function getSubdomain()
+    protected function getSubdomain()
     {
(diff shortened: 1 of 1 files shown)


atymic:
Usually the best way to do this is is to just override the service container, like so:
https://github.com/SocialiteProviders/Providers/tree/master/src/Azure#advanced-usage

Does this not work for your use case?

27pchrisl:
I could do for this specific instance, but I normally override the provider because there's something I need to change like the scopes, refresh token behaviour, fields in the URL etc. The functions that enable this like getScopes, getCodeFields, getTokenFields are all protected. This is the only one that I've wanted to change and it's private.

PR #1311, opened 2024-12-30
Fix: diffrent request resposne from diffrent hosts

realpoke:
This pull request addresses an inconsistency in OpenID request responses observed between local development and AWS Lambda deployments. 

- **Normalization:** The code now normalizes OpenID request keys by consistently using the `openid.` previously it was sometimes `openid_` and other times`openid.` depending on where the code was executed. This ensures consistent handling across different environments.
- **Constant Definitions:** Introduced constants for common OpenID keys (`self::OPENID_RETURN_TO`, `self::OPENID_CLAIMED_ID`) to improve code readability, maintainability, and reduce the risk of typos.
- **Improved Error Handling:** Replaced the generic "unknown error" return with a more specific `OpenIDValidationException` when critical OpenID parameters are missing from the request, enhancing debugging and error handling.

This change ensures more robust and predictable behavior of the OpenID integration across different environments.

### DIFF
--- a/src/Steam/Provider.php
+++ b/src/Steam/Provider.php
@@ -219,15 +219,9 @@ class Provider extends AbstractProvider
      */
     private function normalizeOpenidKeys()
     {
-        $normalized = [];
-        foreach ($this->request->all() as $key => $value) {
-            if (str_starts_with($key, 'openid_')) {
-                $normalizedKey = 'openid.'.substr($key, 7);
-            } else {
-                $normalizedKey = $key;
-            }
-            $normalized[$normalizedKey] = $value;
-        }
+        $normalized = $this->request->collect()->mapWithKeys(function ($value, $key) {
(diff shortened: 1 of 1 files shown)


SvenningsenDev:
lgtm

src/Steam/Provider.php
```diff
@@ -200,6 +212,26 @@ private function requestIsValid()
...
     }
 
+    /**
+     * Normlize openid keys from diffrent requests
+     *
+     * @return void
+     */
+    private function normalizeOpenidKeys()
+    {
+        $normalized = [];
```
lucasmichot, line 222: Why not just:
```php
$normalized = $this->request->collect()->mapWithKeys(function ($value, $key) {
    return [preg_replace('/^openid_/', 'openid.', $key) => $value];
})->all();
```

realpoke: Could work too.

atymic (approved):

PR #1312, opened 2025-01-02
[PropelAuth] support php 7.4

pfvatterott:


atymic:
7.4 is way out of support. You can fork and manually change if you need

PR #1315, opened 2025-01-03
Require php 8.1 and manager 4.8

lucasmichot:
Require php 8.1 and manager 4.8


atymic (approved):
@lucasmichot happy for you to merge this, however needs minor version tags for every repo. Maybe worth writing a tool like `tools/release-major-version.php` as its gonna be a big effort and i don't have time atm.

PR #1318, opened 2025-01-03
use additionalConfigKeys property

lucasmichot:
Direct use `$additionalConfigKeys` property when possible


atymic:
Not sure this bring any benefits, and we also lose git blame history.

PR #1331, opened 2025-02-06
Add LlaveMX OAuth Provider to Laravel Socialite

GonzaloMTZRZ:
This pull request introduces support for LlaveMX as an OAuth provider. LlaveMX is a government-backed authentication system in Mexico.


GonzaloMTZRZ:
I am closing this pull request because we have implemented a private Composer package manager in our government dependency, and this change is no longer aligned with the current configuration. Thank you for the effort put into this PR.

Issue #1308, opened 2024-12-13
Clerk Mapping from nickname to username throws error


LTorrealbaUpstart13:
While using clerk, the user() object mpas username to nickname, it seems like username is no longer valid and throws an error

atymic:
Please add the exception, and the raw array data so we can fix, or PR a fix yourself.

rajdeut:
![Image](https://github.com/user-attachments/assets/e3d01ea9-ae9f-4c65-b78a-ad2028a7c455)

The returned user from Clerk has no `username` key unless it's set in Clerk. When it's empty or usernames are disabled in Clerk the key is missing and it throws an exception ```Undefined array key "username"```

Will submit a PR to update the `mapUserToObject` function to:
```    protected function mapUserToObject(array $user)
    {
        return (new User)->setRaw($user)->map([
            'id' => $user['user_id'],
            'nickname' => $user['username'] ?? ($user['given_name'] ?? explode('@', $user['email'])[0]),
            'name' => $user['name'],
            'email' => $user['email'],
            'avatar' => $user['picture'],
        ]);
    }```

atymic:
Closed in #1344

PR #1344, opened 2025-03-10
fix: Clerk failing to map nickname when username empty

rajdeut:
Username isn't always returned by Clerk and this is failing. Check for username to map otherwise use given_name or as fallback email address.

Issue: https://github.com/SocialiteProviders/Providers/issues/1308

### DIFF
--- a/src/Clerk/Provider.php
+++ b/src/Clerk/Provider.php
@@ -68,11 +68,11 @@ class Provider extends AbstractProvider
     protected function mapUserToObject(array $user)
     {
         return (new User)->setRaw($user)->map([
-            'id'       => $user['user_id'],
-            'nickname' => $user['username'],
-            'name'     => $user['name'],
-            'email'    => $user['email'],
-            'avatar'   => $user['picture'],
+            'id' => $user['user_id'],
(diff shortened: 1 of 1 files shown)


src/Clerk/Provider.php
```diff
@@ -68,11 +68,11 @@ protected function getUserByToken($token)
     protected function mapUserToObject(array $user)
     {
         return (new User)->setRaw($user)->map([
-            'id'       => $user['user_id'],
-            'nickname' => $user['username'],
-            'name'     => $user['name'],
-            'email'    => $user['email'],
-            'avatar'   => $user['picture'],
+            'id' => $user['user_id'],
+            'nickname' => $user['username'] ?? ($user['given_name'] ?? explode('@', $user['email'])[0]),
```
atymic, line 72: Can we drop the email fallback? can just return null in that case

atymic (changes requested):

src/Clerk/Provider.php
rajdeut: Done.

atymic (approved):

PR #1345, opened 2025-03-13
Add NFDI Login provider

mzur:
This is the authentication method for the German national research data infrastructure.

See: https://nfdi-aai.de/


mzur:
Ping

atymic (approved):

PR #1346, opened 2025-03-13
Support Threads users without profile pictures

grrrian:
This prevents a crash when a user doesn't have an avatar.

Without this we experience this error:

```
ERROR: Error handling Threads callback {"exception":"Undefined array key \"threads_profile_picture_url\"","stack_trace":"#0 /[REDACTED]/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(255): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'Undefined array...', '[REDACTED]', 84)\n#1 [REDACTED]/vendor/socialiteproviders/threads/Provider.php(84)
```

### DIFF
--- a/src/Threads/Provider.php
+++ b/src/Threads/Provider.php
@@ -73,9 +73,7 @@ class Provider extends AbstractProvider
             'nickname'     => Arr::get($user, 'username'),
             'name'         => null,
             'email'        => null,
-            'avatar'       => array_key_exists('threads_profile_picture_url', $user)
-                                ? $user['threads_profile_picture_url']
-                                : null,
+            'avatar'       => Arr::get($user, 'threads_profile_picture_url', null),
         ]);
     }
 }


atymic:
The third param is optional, this will already return null:
Screen Shot 2025-03-18 at 18 02 44@2x

grrrian:
My bad. Closing PR. It turns out I had this issue because my code didn't have this safe getter; for some reason my code pre-dated this [PR](https://github.com/SocialiteProviders/Providers/commit/ee94f0be0ba861c07c89bc1594bf606d152c8eff) which is strange because I installed it using the `composer require socialiteproviders/threads` command about a month ago 🤔 

Out of curiosity, does that command not fetch the latest version?

atymic:
Huh, that's strange. Potentially there was a lag in the composer release. Glad it's fixed tho :)

PR #1348, opened 2025-03-20
Add PixelTools provider

TeaRiot:


src/PixelTools/PixelToolsExtendSocialite.php
```diff
@@ -0,0 +1,13 @@
...
+
+use SocialiteProviders\Manager\SocialiteWasCalled;
+
+class PixelToolsExtendSocialite
+{
+    public function handle(SocialiteWasCalled $socialiteWasCalled): void
+    {
+        $socialiteWasCalled->extendSocialite('pixeltools', Provider::class);
+    }
+}
```
atymic, line 13: missing newline

src/PixelTools/composer.json
```diff
@@ -0,0 +1,33 @@
...
+    "php": "^8.0",
+    "ext-json": "*",
+    "socialiteproviders/manager": "^4.4"
+  },
+  "autoload": {
+    "psr-4": {
+      "SocialiteProviders\\PixelTools\\": ""
+    }
+  }
+}
```
atymic, line 33: Same here

atymic (changes requested):
Can you let me know what this site does? Not sure it needs to be a public provider.

PR #1349, opened 2025-03-24
feat: Add Didit provider

alexstewartja:
Docs: https://docs.didit.me/auth-and-data/sign-in-api-reference/full-flow


alexstewartja:
@atymic Hey, I can tell you're probably busy. If you need help with maintenance/review, I'm available to chip in when I can.

PR #1350, opened 2025-04-19
feat: VK ID OAuth2 Provider

vhar:
New VK ID OAuth2 Provider

[OAuth VKontakte](https://github.com/SocialiteProviders/Providers/tree/master/src/VKontakte) now is deprecated.  
Obtaining a user access key (access token) is now possible only through OAuth VK ID.   

[VK ID Reference](https://id.vk.com/about/business/go/docs/ru/vkid/latest/oauth-vk#OAuth-VKontakte)

### DIFF
--- a/src/VKID/Provider.php
+++ b/src/VKID/Provider.php
@@ -78,29 +78,6 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
-    public function user()
-    {
-        if ($this->hasInvalidState()) {
-            throw new InvalidStateException();
-        }
-
-        $response = $this->getAccessTokenResponse($this->getCode());
-
-        $user = $this->mapUserToObject($this->getUserByToken($response));
-
-        $this->credentialsResponseBody = $response;
-
-        if ($user instanceof User) {
-            $user->setAccessTokenResponseBody($this->credentialsResponseBody);
-        }
-
-        return $user->setToken($this->parseAccessToken($response))
-            ->setExpiresIn($this->parseExpiresIn($response));
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     protected function mapUserToObject(array $user)
     {
         return (new User())->setRaw($user)->map([
--- a/src/VKID/Provider.php
+++ b/src/VKID/Provider.php
@@ -1,0 +1,134 @@
+<?php
+
+namespace SocialiteProviders\VKID;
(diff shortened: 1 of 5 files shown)


Ichinya:
That's exactly what I need.

atymic:
Is there a reason why this can't just replace the old provider with a new major version?

vhar:
> Is there a reason why this can't just replace the old provider with a new major version?

Yes, there is. VK applications created earlier continue to work with the old provider.
OAuth VKontakte
This is an outdated way of authorization using OAuth VKontakte. Getting a user Access token is now only possible through OAuth VK ID. At the same time, previously created applications with OAuth VKontakte are supported.
If you need to work with the API on behalf of a community or public page, for example, so that the user can respond to community subscribers to messages, get a list of administered community identifiers through the user's [Access token via OAuth 2.1](https://id.vk.com/about/business/go/docs/en/vkid/latest/vk-id/connection/tokens/access-token), and then with this information get a community access key via OAuth VKontakte.
https://id.vk.com/about/business/go/docs/en/vkid/latest/oauth-vk#OAuth-VKontakte

src/VKID/Provider.php
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function user()
+    {
+        if ($this->hasInvalidState()) {
+            throw new InvalidStateException();
+        }
+
+        $response = $this->getAccessTokenResponse($this->getCode());
```
atymic, line 87: Is there a reason why this is overriden?

src/VKID/composer.json
```diff
@@ -0,0 +1,34 @@
...
+    ],
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Vladimir Kharinenkov",
+            "email": "vhar@mail.ru"
+        }
+    ],
+    "require": {
+        "php": "^7.4 || ^8.0",
```
atymic, line 20: non php 7, lets limit to 8.4+

atymic (changes requested):

atymic (approved):

src/VKID/composer.json
atymic: https://endoflife.date/php
We can move to 8.2, but all other are basically out of sec support, and long out of active maintenance

Issue #1354, opened 2025-05-02
Timezone error in apple login .


harrowmykel:
in \socialiteproviders\apple\Provider.php: 155
            ```php
             $constraints = [
                new SignedWith(new Sha256(), AppleSignerInMemory::plainText($publicKey['key'])),
                new IssuedBy(self::URL),

                // the culprit >>
                new LooseValidAt(SystemClock::fromSystemTimezone()),
            ];
            try {
                $jwtContainer->validator()->assert($token, ...$constraints);

                return true;
            } catch (RequiredConstraintsViolated $e) {
                throw new InvalidStateException($e->getMessage());
            }
            ```
            the class LooseValidAt() checks that a token was not issued in the future by checking
            that iat is at least `PT0S` in the past. this causes a problem if the timezone is different 
            or if the token is immediately generated by apple and the difference is less than 0 seconds due to different timezone.
            how can we make leeway value adjustable? or fix this
            ```php
            private function assertIssueTime(Token $token, DateTimeInterface $now): void
            {
                if (! $token->hasBeenIssuedBefore($now)) {
                    throw ConstraintViolation::error('The token was issued in the future', $this);
                }
            }
            ```

harrowmykel:
Example of the code and the result. As you can see the value of 
`(SystemClock::fromSystemTimezone())->now()` is greater than ` $token->claims()->get(RegisteredClaims::ISSUED_AT),`
by 0.27 seconds

![Image](https://github.com/user-attachments/assets/b8519a4d-d3ba-4af0-8e26-9cff1b4d5a34)

![Image](https://github.com/user-attachments/assets/19f1b123-99c9-4a36-9b57-4789138c9836)

harrowmykel:
it is currently not possible to fix this issue in the providers package since class LooseValidAt does not accept a negative leeway.

I have created an issue now at JWT
https://github.com/lcobucci/jwt/issues/1110

SvenRtbg:
Just to state it here: This isn't an issue with time zones. JWT are based on unix timestamps, which doesn't know about time zones, and PHP DateTime classes are well suited to handle time zones correctly.

[...]

PR #1355, opened 2025-05-07
Improve lcobucci/jwt usage in Apple provider

SvenRtbg:
Using the Configuration class forces to know a signer and a key, which is irrelevant for the purpose of this provider, but it introduces some code that looks scary if used incorrectly.

The irrelevant classes are removed, additional exception cases are now rethrowing as InvalidStateException. Conditions have been inverted to implement early exit.


src/Apple/AppleSignerNone.php
SvenRtbg, line 1: This file suggests it might be used to implement the "none" token signature, which is known to create security issues, as "none" is a value that is attacker-provided and might be used to trick vulnerable JWT validators in ignoring the signature.

I have not investigated if this class could be abused for such a scenario, but it looks scary.

src/Apple/AppleSignerInMemory.php
SvenRtbg, line 1: This class is a copy of the `InMemory` class from lcobucci/jwt with the "empty key" check removed. It is used to get an instance of the `Configuration` class without needing to provide a key that isn't used at all.

src/Apple/Provider.php
```diff
@@ -119,11 +122,11 @@ public function userByIdentityToken(string $token): User
      */
     public static function verify($jwt)
     {
-        $jwtContainer = Configuration::forSymmetricSigner(
-            new AppleSignerNone,
-            AppleSignerInMemory::plainText('')
-        );
-        $token = $jwtContainer->parser()->parse($jwt);
```
SvenRtbg, line 126: I know the documentation from lcobucci/jwt talks about this `Configuration` object for easier use, but the only things it needs to do for the current use case is to provide a parser and a validator.

The parser only has a simple depencency on the `JoseEncoder` class, the validator is free of other dependencies. So it does not make sense to deal with providing a signer for "none" (explained above that this is scary because it might be insecure), and a `Key` class with an empty key that are never going to be used for token validation.

src/Apple/Provider.php
```diff
@@ -134,24 +137,23 @@ public static function verify($jwt)
         $publicKeys = JWK::parseKeySet($data);
         $kid = $token->headers()->get('kid');
 
-        if (isset($publicKeys[$kid])) {
-            $publicKey = openssl_pkey_get_details($publicKeys[$kid]->getKeyMaterial());
+        if (!isset($publicKeys[$kid])) {
```
SvenRtbg, line 140: This `if` is inverted to allow shortcut throwing the exception instead of dealing with an entire indentation level for the good case.

src/Apple/Provider.php
```diff
@@ -134,24 +137,23 @@ public static function verify($jwt)
...
+            throw new InvalidStateException('Invalid JWT Signature');
+        }
+
+        $publicKey = openssl_pkey_get_details($publicKeys[$kid]->getKeyMaterial());
+        try {
             $constraints = [
-                new SignedWith(new Sha256, AppleSignerInMemory::plainText($publicKey['key'])),
+                new SignedWith(new Sha256, InMemory::plainText($publicKey['key'])),
```
SvenRtbg, line 147: Switching from `AppleSignerInMemory` to `InMemory`, which provides checks against empty keys, and has (in more recent versions) also annotations for `#[SensitiveParameter]` to hide secrets.

SvenRtbg:
This is vaguely related to https://github.com/SocialiteProviders/Providers/issues/1354, triggered by the sub-issue https://github.com/lcobucci/jwt/issues/1110 which made me look at this code.

Still the question from #1354 remains whether or not it might be possible to configure a leeway when comparing timestamps, as the current implementation expects clocks to be perfectly in sync, especially when consuming the Apple token on a server that is a bit behind the current time and takes offense that the token was created in the future (even if only by half a second),

SvenRtbg:
Who needs to be bugged from the maintainers to give some feedback whether or not you want to merge this? @atymic maybe?

src/Apple/Provider.php
```diff
@@ -114,43 +114,53 @@ protected function getUserByToken($token)
...
+        if (!$this->jwtConfig instanceof Configuration) {
             $private_key_path = $this->getConfig('private_key', '');
             $private_key_passphrase = $this->getConfig('passphrase', '');
-            $signer = $this->getConfig('signer', '');
+            $signerClassName = $this->getConfig('signer', '');
 
-            if (empty($signer) || !class_exists($signer)) {
-                $signer = !empty($private_key_path) ? \Lcobucci\JWT\Signer\Ecdsa\Sha256::class : AppleSignerNone::class;
+            if (empty($signerClassName) || !class_exists($signerClassName) || !is_a($signerClassName, Signer::class, true)) {
+                $signerClassName = EcdsaSha256::class;
```
SvenRtbg, line 133: The `AppleSignerNone` is gone.

I don't believe there is a use case for a configurable signer class, as the only signature allowed by Apple is ES256 right now, but I added some sanity check beyond "class exists".

atymic:
@SvenRtbg can you foresee any breaking changes? Thinking of releasing a beta so people can test first.

SvenRtbg:
@atymic Obviously any usage of the removed classes will lead to failure, but I understood them as being internal with no intentional public exposure.

The detailed usage of secret vs. private key change is something I cannot assess at all, but it is already introduced in `master` anyways, and I don't see how I would have destroyed that (fingers crossed) - or introduced any breaking changes. 

Initially I was trying to clean up validating an Apple token - creating one for authentication purposes came as an unexpected feature. :)

PR #1361, opened 2025-05-20
feat: add roles claim to Microsoft Provider 

akadoya:
Adding roles attribute so that the client application can consume it and assign users to roles at login. 
(or...exposing id_token itself also works like Okta Provider does, if you'd prefer that way)

### DIFF
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -156,6 +156,8 @@ class Provider extends AbstractProvider
             $formattedResponse['tenant'] = json_decode((string) $responseTenant->getBody(), true)['value'][0] ?? null;
         }
 
+        $formattedResponse['id_token'] = $this->credentialsResponseBody['id_token'] ?? null,
+
         return $formattedResponse;
     }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft/Provider.php
```diff
@@ -156,6 +156,8 @@ protected function getUserByToken($token)
             $formattedResponse['tenant'] = json_decode((string) $responseTenant->getBody(), true)['value'][0] ?? null;
         }
 
+        $formattedResponse['roles'] = $this->getRoles();
```
atymic, line 159: Can we please get an example in the readme docs of the format of this please

atymic (changes requested):

src/Microsoft/Provider.php
akadoya: @atymic The returned role names' format would depend on how you configure the Entra ID side... but I updated README from what I know....would this work?

akadoya: Gotcha, it returns an array of string for user group names. updated readme and also line 233 accordingly.

atymic (approved):

PR #1367, opened 2025-06-19
feat: add Crowdin provider

WentTheFox:
Implemented by duplicating the Discord provider and following this service-specific guide: https://support.crowdin.com/developer/authorizing-oauth-apps/

The Crowdin OAuth API requires a scope to be specified, otherwise the authorization flow breaks. The list of available scopes is at https://support.crowdin.com/developer/understanding-scopes/ but there does not appear to be a generic "just the basic user info please" scope. "Translation status" (`project.status`) is one of, if not the least invasive from that list which also does not require an Enterprise license, that's why I opted to make it the default.

Side note: when using the default Laravel `php artisan serve` command, which uses `127.0.0.1:8000` as the host, authorization requests get mysteriously rejected with a 403 error by the service when this is used in a `redirect_uri`. For my testing I had to change things around to use `localhost:8000` in the `APP_URL` env var and had to run `php artisan serve --host=localhost` to get this to work. This would not affect a real deployment behind an actual URL, but I'm not sure it's worth noting this anywhere.

### DIFF
--- a/monorepo-builder.yml
+++ b/monorepo-builder.yml
@@ -39,6 +39,7 @@ parameters:
     src/Coinbase: 'git@github.com:SocialiteProviders/Coinbase.git'
     src/ConstantContact: 'git@github.com:SocialiteProviders/ConstantContact.git'
     src/Coursera: 'git@github.com:SocialiteProviders/Coursera.git'
+    src/Crowdin: 'git@github.com:SocialiteProviders/Crowdin.git'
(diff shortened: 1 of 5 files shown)


atymic (approved):
Thank you :)
Please PR to website
https://github.com/SocialiteProviders/website

WentTheFox:
@atymic Opened here: https://github.com/SocialiteProviders/website/pull/76

PR #1368, opened 2025-06-20
Fix: code_challenge error

dominikhosic:
Upon OAuth login on TikTok, there was an error related to ‘code_challenge.’ I found out that this logic is not implemented in the provider at all.


atymic (changes requested):
Is there a reason why you can't use ` protected $usesPKCE = true;`?

atymic:
ping @dominikhosic

PR #1393, opened 2025-09-16
Fix for Mailru Provider

baglan1:
The "nickname" field, in a user object, is not always returned since it's a relatively new field. Therefore, old mail.ru accounts do not have that field and mail.ru OAuth does not return that field in their response.

This bug breaks mail.ru OAuth for older accounts and some users cannot register or login via Mailru Provider.

### DIFF
--- a/src/Mailru/Provider.php
+++ b/src/Mailru/Provider.php
@@ -43,7 +43,7 @@ class Provider extends AbstractProvider
     {
         return (new User)->setRaw($user)->map([
             'id'       => $user['id'],
-            'nickname' => $user['nickname'],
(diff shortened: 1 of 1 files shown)


src/Mailru/Provider.php
```diff
@@ -43,7 +43,7 @@ protected function mapUserToObject(array $user)
     {
         return (new User)->setRaw($user)->map([
             'id'       => $user['id'],
-            'nickname' => $user['nickname'],
+            'nickname' => \Arr::get($user, 'nickname'),
```
atymic, line 46: can we import Arr please

atymic (approved):

PR #1397, opened 2025-10-06
added the ability to use the pkce

guilebc:
I have added the option to use PKCE. TikTok requires this feature for local use (https://developers.tiktok.com/doc/login-kit-desktop).


atymic (changes requested):
Can't we just set $usePKCE?

Issue #1405, opened 2025-11-13
Add support for configurable RelayState in SAML2 provider


biswajit287:
### **Description**

I’m using the SAML2 provider to implement Single Sign-On (SSO) with Microsoft Entra ID.

Currently, there’s no way to configure or control the RelayState value in SP-initiated SSO requests, which causes issues when redirecting users back to their intended URL after authentication.

### **Problem**

There are two main scenarios where this limitation causes problems:

Session timeout

> When a user’s session expires and they are redirected to the login page, the intended URL is lost after re-login.

Shared links

> When a user opens a shared deep link that requires authentication, they are redirected to the login page. After logging in, they are not redirected back to the original URL, since the RelayState is not preserved.

### From Microsoft’s documentation:

> The value in this form only takes effect in an IdP-initiated SSO flow.
> If your application uses SP-initiated SSO, then the service provider needs to send the RelayState as a parameter in the SAML request.
> 
> The RelayState instructs the application where to redirect users after authentication is completed, and the value is typically a URL or URL path that takes users to a specific location within the application.

### **Root Cause**

I found that the provider sets a random RelayState value by default and doesn’t allow user configuration.

vendor/socialiteproviders/saml2/Provider.php
```
if ($this->usesState()) {
    $this->request->session()->put('state', $state = $this->getState());
    $authnRequest->setRelayState($state);
}
```

vendor/laravel/socialite/src/Two/AbstractProvider.php

```
protected function getState()
{
    return Str::random(40);
}

```

### **Proposed Solution**
Allow developers to define a custom RelayState when initiating the SAML login.

Example approach:

- Override getState() in the provider to check if a relayState parameter exists in $parameters.
- If present, use that value instead of generating a random string.
- Fallback to the default random value when relayState is not provided.

This would enable redirecting users back to the intended page after SSO login (e.g., post-session timeout or shared deep links).

If there’s an existing way to achieve this with the current implementation, I’d appreciate any guidance.
Otherwise, I’d be happy to open a PR to add this feature.

biswajit287:
Hi, any update on this? Should I create a PR addressing this issue?

PR #1411, opened 2025-12-15
use RelayState from parameters when present as the state, fallback to default otherwise.

biswajit287:
This PR changes behavior so that only explicitly provided RelayState values are used as state. Missing or empty values now fall back to the default state generation.

For more info checkout this [issue](https://github.com/SocialiteProviders/Providers/issues/1405)


atymic:
Thanks @27pchrisl and @biswajit287

biswajit287:
Hi @atymic, thanks for merging the PR. Do you have an estimate for when this will be released?

PR #1414, opened 2026-01-01
feat: Add Kick OAuth2 provider

itsalifadaei:
### Added Kick OAuth2 Provider

This PR adds a new Socialite provider for Kick.

- OAuth2 implementation
- Follows SocialiteProviders standards
- Includes documentation and docgen updates

### DIFF
--- a/src/Kick/Provider.php
+++ b/src/Kick/Provider.php
@@ -49,7 +49,7 @@ class Provider extends AbstractProvider
      */
     protected function mapUserToObject(array $user)
     {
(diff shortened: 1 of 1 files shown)


KowalskyCode:
It appears that Kick uses OAuth 2.1, which requires PKCE for the authorization flow. I was encountering errors until I enabled it. It seems we are missing protected `$usesPKCE= true;` in the Provider class. Adding this fixed the issue for me.

atymic (approved):

PR #1415, opened 2026-01-03
bugfix: updated classname

wivaku:


wivaku:
Would it be possible to merge this one?

PR #1422, opened 2026-01-22
Use config from Provider instead of config(‘services.apple’)

bonroyage:
Rather than retrieving the `team_id`, `client_id` and `key_id` from the config file, this uses the config that is known by the Provider at the time the AppleToken is created.

In our scenario, we dynamically build the config (as explained here: https://socialiteproviders.com/usage/#overriding-a-config) and thus don't have `services.apple` configured.

I realise it is a breaking change to introduce new required parameters. If preferred, I can make the parameters optional and read from the config like before if they're omitted.

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -365,6 +365,6 @@ class Provider extends AbstractProvider
      */
     public static function additionalConfigKeys()
     {
-        return ['private_key', 'passphrase', 'signer'];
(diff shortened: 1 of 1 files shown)


bonroyage:
> Yes, please make sure it's BC compat change :)

Done!

atymic:
Have you tested this in your APP? 
I don't think this will work as the service container isn't going to resolve those params - as per the example you linked they need to be in the `$additionalProviderConfig` array

PR #1444, opened 2026-03-25
Fix Microsoft OpenID configuration cache serialization

ubertech-za:
getOpenIdConfiguration() caches json_decode output as stdClass. On deserialization from file/database cache drivers, PHP returns __PHP_Incomplete_Class, breaking subsequent login attempts.

Fix: return associative arrays instead of stdClass objects, and update all property access to array syntax.


atymic (changes requested):
@ubertech-za can you please change the cache key so its different and does not result in errors for existing users :)

atymic (approved):

Issue #823, opened 2022-03-30
OpenID Connect implementation


fabricecw:
There are multiple issues about the implementation of OpenID Connect. In #751 it seems OpenID is not yet on the roadmap. What do you think about a generic OpenID Connect provider? If you think it makes sense, we will take care of it and PR this provider.

Braunson:
Following up on this!

fabricecw:
I'm still confused why no one seems to be interested in an OpenID Connect implementation. OIDC became a popular standard when it comes to SSO authentication. At least for enterprises.

Maybe I am missing something.

PR #1447, opened 2026-04-23
Add OpenID Connect

adrum:
This adds an [OpenID Connect](https://openid.net/connect/) provider to make it easier to integrate a wide range of supported providers. It implements the [Authorization Code Flow ](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) with [PKCE](https://datatracker.ietf.org/doc/html/rfc7636) by default, [auto-discovery](https://openid.net/specs/openid-connect-discovery-1_0.html) via .well-known/openid-configuration (which is cached by default with configurable TTL), JWKS signature verification with kid-miss invalidation for real-time key rotation, configurable signing algorithms, and full spec-level validation of iss, aud, azp, at_hash, nonce, and exp/nbf/iat (with configurable clock-skew leeway) in both verified and unverified paths. 

I also provided convenient ways to handle logout() for [RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html), revoke() for [RFC 7009](https://datatracker.ietf.org/doc/html/rfc7009) token revocation, and verifyLogoutToken() for [Back-Channel Logout 1.0](https://openid.net/specs/openid-connect-backchannel-1_0.html) with complete claim validation. 

The README documents setup, every config key, and working examples for session- and database-backed token storage plus a sid → Laravel session ID mapping table so we can individually clear the correct session.

Completes this stale closed issue: https://github.com/SocialiteProviders/Providers/issues/823

### DIFF
--- a/src/OpenIDConnect/Provider.php
+++ b/src/OpenIDConnect/Provider.php
@@ -1,0 +1,538 @@
+<?php
+
+namespace SocialiteProviders\OpenIDConnect;
+
+use Exception;
+use Firebase\JWT\JWK;
+use Firebase\JWT\JWT;
+use Firebase\JWT\Key;
+use GuzzleHttp\Exception\GuzzleException;
+use GuzzleHttp\RequestOptions;
+use Illuminate\Http\RedirectResponse;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Str;
+use InvalidArgumentException;
+use JsonException;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    public const IDENTIFIER = 'OPENIDCONNECT';
+
+    public $configurations = null;
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = [
+        'openid',
+        'email',
+        'profile',
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    /**
+     * Indicates if the nonce should be utilized.
+     */
+    protected bool $usesNonce = true;
+
+    /**
+     * Use PKCE by default (Authorization Code Flow + PKCE).
+     * This follows OAuth 2.1 / current OIDC best practice.
+     */
+    protected $pkceEnabled = true;
+
+    /**
+     * Indicates if JWT signature verification should be enabled.
+     * Can be overridden by config 'verify_jwt'.
+     */
+    protected bool $verifyJwt = false;
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function additionalConfigKeys(): array
+    {
+        return [
+            'base_url',
+            'scopes',
+            'verify_jwt',
+            'jwt_public_key',
+            'jwt_algorithm',
+            'issuer',
+            'token_auth_method',
+            'post_logout_redirect_uri',
+        ];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function redirect(): RedirectResponse
+    {
+        $state = null;
+
+        if ($this->usesState()) {
+            $this->request->session()->put('state', $state = $this->getState());
+        }
+
+        if ($this->usesNonce()) {
+            $this->request->session()->put('nonce', $this->getNonce());
+        }
+
+        if ($this->usesPKCE()) {
+            $this->request->session()->put('code_verifier', $this->getCodeVerifier());
+        }
+
+        return new RedirectResponse($this->getAuthUrl($state));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getScopes(): array
+    {
+        if ($this->getConfig('scopes')) {
+            return array_merge($this->scopes, explode(' ', $this->getConfig('scopes')));
+        }
+
+        return $this->scopes;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getTokenUrl(): string
+    {
+        return $this->getOpenIdConfig()['token_endpoint'];
+    }
+
+    /**
+     * Get the userinfo URL for the provider.
+     *
+     * @throws GuzzleException
+     */
+    protected function getUserInfoUrl(): string
+    {
+        return $this->getOpenIdConfig()['userinfo_endpoint'];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getAuthUrl($state): string
+    {
+        return $this->buildAuthUrlFromBase(
+            $this->getOpenIdConfig()['authorization_endpoint'],
+            $state
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getCodeFields($state = null): array
+    {
+        $fields = parent::getCodeFields($state);
+
+        if ($this->usesNonce()) {
+            $fields['nonce'] = $this->getCurrentNonce();
+        }
+
+        return $fields;
+    }
+
+    /**
+     * Determine if the provider is operating with nonce.
+     */
+    protected function usesNonce(): bool
+    {
+        return $this->usesNonce;
+    }
+
+    /**
+     * Get a newly generated nonce.
+     */
+    protected function getNonce(): string
+    {
+        return Str::random(40);
+    }
+
+    /**
+     * Determine if JWT signature verification is enabled.
+     */
+    protected function shouldVerifyJwt(): bool
+    {
+        return (bool) $this->getConfig('verify_jwt', $this->verifyJwt);
+    }
+
+    /**
+     * Get the current nonce stored in the session.
+     */
+    protected function getCurrentNonce(): ?string
+    {
+        return $this->request->session()->get('nonce');
+    }
+
+    /**
+     * @throws GuzzleException
+     */
+    protected function getOpenIdConfig(): array
+    {
+        if ($this->configurations === null) {
+            $configUrl = rtrim($this->getConfig('base_url'), '/').'/.well-known/openid-configuration';
+            $cacheKey = 'openidconnect_discovery_'.md5($configUrl);
+
+            $this->configurations = Cache::remember($cacheKey, 3600, function () use ($configUrl) {
+                try {
+                    $response = $this->getHttpClient()->get($configUrl);
+
+                    return json_decode((string) $response->getBody(), true, 512, JSON_THROW_ON_ERROR);
+                } catch (Exception $e) {
+                    throw new InvalidArgumentException('Unable to get the OIDC configuration from '.$configUrl.': '.$e->getMessage());
+                }
+            });
+        }
+
+        return $this->configurations;
+    }
+
+    /**
+     * Get the JSON Web Key Set from the OIDC provider.
+     *
+     * @throws GuzzleException
+     */
+    protected function getJwks(): array
+    {
+        $cacheKey = 'openidconnect_jwks_'.md5($this->getConfig('base_url'));
+
+        return Cache::remember($cacheKey, 3600, function () {
+            $config = $this->getOpenIdConfig();
+
+            if (! isset($config['jwks_uri'])) {
+                throw new InvalidArgumentException('JWKS URI not found in OIDC configuration');
+            }
+
+            try {
+                $response = $this->getHttpClient()->get($config['jwks_uri']);
+
+                return json_decode((string) $response->getBody(), true, 512, JSON_THROW_ON_ERROR);
+            } catch (Exception $e) {
+                throw new InvalidArgumentException('Unable to fetch JWKS: '.$e->getMessage());
+            }
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function user()
+    {
+        if ($this->user) {
+            return $this->user;
+        }
+
+        if ($this->hasInvalidState()) {
+            throw new InvalidArgumentException('Callback: invalid state.', 401);
+        }
+
+        $tokenResponse = $this->getAccessTokenResponse($this->request->input('code'));
+
+        $payload = $this->decodeJWT($tokenResponse['id_token'], $tokenResponse['access_token'] ?? null);
+
+        if ($this->hasEmptyEmail($payload)) {
+            $payload = $this->getUserByToken($tokenResponse['access_token']);
+            if (empty($payload['email'] ?? null)) {
+                throw new InvalidArgumentException('JWT: User has no email.', 401);
+            }
+        }
+
+        $raw = (array) $payload;
+        $raw['id_token'] = $tokenResponse['id_token'];
+
+        $this->user = $this->mapUserToObject($raw);
+
+        return $this->user->setToken($tokenResponse['access_token'])
+            ->setRefreshToken($tokenResponse['refresh_token'] ?? null)
+            ->setExpiresIn($tokenResponse['expires_in']);
+    }
+
+    protected function decodeJWT(string $jwt, ?string $accessToken = null)
+    {
+        $header = $this->decodeJwtHeader($jwt);
+        $alg = $header->alg ?? null;
+
+        if ($this->shouldVerifyJwt()) {
+            $payload = $this->verifyAndDecodeJWT($jwt, $alg);
+        } else {
+            try {
+                [, $jwtPayload] = explode('.', $jwt);
+                $payload = json_decode($this->base64UrlDecode($jwtPayload));
+            } catch (Exception $e) {
+                throw new InvalidArgumentException('JWT: Failed to parse.', 401);
+            }
+        }
+
+        $this->validateIdTokenClaims($payload, $alg, $accessToken);
+
+        if ($this->usesNonce()) {
+            $this->request->session()->forget('nonce');
+        }
+
+        return $payload;
+    }
+
+    /**
+     * Verify the JWT signature and decode it.
+     */
+    protected function verifyAndDecodeJWT(string $jwt, ?string $alg)
+    {
+        try {
+            $publicKey = $this->getConfig('jwt_public_key');
+            $configuredAlg = $this->getConfig('jwt_algorithm') ?: $alg ?: 'RS256';
+
+            if ($publicKey) {
+                $decoded = JWT::decode($jwt, new Key($publicKey, $configuredAlg));
+            } else {
+                $decoded = JWT::decode($jwt, JWK::parseKeySet($this->getJwks(), $configuredAlg));
+            }
+
+            return json_decode(json_encode($decoded));
+        } catch (Exception $e) {
+            throw new InvalidArgumentException('JWT: Verification failed - '.$e->getMessage(), 401);
+        }
+    }
+
+    protected function decodeJwtHeader(string $jwt)
+    {
+        try {
+            [$headerB64] = explode('.', $jwt);
+
+            return json_decode($this->base64UrlDecode($headerB64));
+        } catch (Exception $e) {
+            throw new InvalidArgumentException('JWT: Failed to parse header.', 401);
+        }
+    }
+
+    /**
+     * Validate the standard OIDC id_token claims: nonce, iss, aud, azp, at_hash.
+     */
+    protected function validateIdTokenClaims($payload, ?string $alg, ?string $accessToken): void
+    {
+        if ($this->isInvalidNonce($payload->nonce ?? null)) {
+            throw new InvalidArgumentException('JWT: Contains an invalid nonce.', 401);
+        }
+
+        $expectedIssuer = $this->getConfig('issuer') ?: ($this->getOpenIdConfig()['issuer'] ?? null);
+        if ($expectedIssuer !== null && ($payload->iss ?? null) !== $expectedIssuer) {
+            throw new InvalidArgumentException('JWT: Invalid issuer.', 401);
+        }
+
+        $aud = $payload->aud ?? null;
+        $audList = is_array($aud) ? $aud : [$aud];
+        if (! in_array($this->clientId, $audList, true)) {
+            throw new InvalidArgumentException('JWT: Invalid audience.', 401);
+        }
+
+        if (is_array($aud) && count($aud) > 1 && ($payload->azp ?? null) !== $this->clientId) {
+            throw new InvalidArgumentException('JWT: Invalid authorized party (azp).', 401);
+        }
+
+        if ($accessToken !== null && isset($payload->at_hash) && $alg) {
+            $this->validateAtHash($payload->at_hash, $accessToken, $alg);
+        }
+    }
+
+    protected function validateAtHash(string $atHash, string $accessToken, string $alg): void
+    {
+        $map = ['256' => 'sha256', '384' => 'sha384', '512' => 'sha512'];
+        $bits = substr($alg, -3);
+
+        if (! isset($map[$bits])) {
+            return;
+        }
+
+        $digest = hash($map[$bits], $accessToken, true);
+        $expected = $this->base64UrlEncode(substr($digest, 0, intdiv(strlen($digest), 2)));
+
+        if (! hash_equals($expected, $atHash)) {
+            throw new InvalidArgumentException('JWT: at_hash mismatch.', 401);
+        }
+    }
+
+    private function base64UrlDecode(string $data): string
+    {
+        return base64_decode(str_pad(strtr($data, '-_', '+/'), strlen($data) % 4, '=', STR_PAD_RIGHT));
+    }
+
+    private function base64UrlEncode(string $data): string
+    {
+        return rtrim(strtr(base64_encode($data), '+/', '-_'), '=');
+    }
+
+    /**
+     * Determine if the current token has a mismatching nonce.
+     */
+    protected function isInvalidNonce($nonce): bool
+    {
+        if (! $this->usesNonce()) {
+            return false;
+        }
+
+        return ! (is_string($nonce) && strlen($nonce) > 0 && $nonce === $this->getCurrentNonce());
+    }
+
+    protected function hasEmptyEmail($payload): bool
+    {
+        if (is_array($payload)) {
+            return empty($payload['email'] ?? null);
+        }
+
+        return empty($payload->email ?? null);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function mapUserToObject(array $user)
+    {
+        return (new User)->setRaw($user)->map([
+            'id'          => $user['sub'] ?? null,
+            'email'       => $user['email'] ?? null,
+            'name'        => $user['name'] ?? null,
+            'nickname'    => $user['nickname'] ?? null,
+            'given_name'  => $user['given_name'] ?? null,
+            'family_name' => $user['family_name'] ?? null,
+            'idp'         => $user['idp'] ?? null,
+            'role'        => $user['role'] ?? null,
+            'groups'      => $user['groups'] ?? null,
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws JsonException|GuzzleException
+     */
+    public function getAccessTokenResponse($code)
+    {
+        $fields = array_merge(
+            $this->getTokenFields($code),
+            ['grant_type' => 'authorization_code']
+        );
+
+        $response = $this->getHttpClient()->post($this->getTokenUrl(), $this->tokenRequestOptions($fields));
+
+        return json_decode((string) $response->getBody(), true, 512, JSON_THROW_ON_ERROR);
+    }
+
+    /**
+     * Build Guzzle request options for the token endpoint, applying the
+     * configured client authentication method (client_secret_post or
+     * client_secret_basic).
+     */
+    protected function tokenRequestOptions(array $fields): array
+    {
+        $method = $this->resolveTokenAuthMethod();
+
+        $options = [RequestOptions::HEADERS => ['Accept' => 'application/json']];
+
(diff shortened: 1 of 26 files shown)


owenvoke:
Just thought I'd mention, I've been running a similar version of this for almost a year in production with no issues. I'd love for this to be natively supported by Socialite Providers!

For context, this is [the version I created for a private app, and copied into my public app a few months ago](https://github.com/owenvoke/keep/blob/main/app/Services/OIDC/SocialiteProvider.php).

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+    /**
+     * Indicates if the nonce should be utilized.
+     */
+    protected bool $usesNonce = true;
+
+    /**
+     * Use PKCE by default (Authorization Code Flow + PKCE).
+     * This follows OAuth 2.1 / current OIDC best practice.
+     */
+    protected $pkceEnabled = true;
```
atymic: **High**

PKCE isn't actually on. Socialite declares `protected $usesPKCE` (Two/AbstractProvider.php:90) and `usesPKCE()` returns that, so `$pkceEnabled` just adds an unused property and the flag stays false. The Manager's `AbstractProvider` in between doesn't touch PKCE at all, and `pkceEnabled` doesn't appear anywhere else in vendor or src.

The rest of the wiring is right, `redirect()` on line 92 and `getCodeFields()` on line 141 both call the real `usesPKCE()`, they're just gated on a flag that never flips. So no `code_challenge` on the authorize request and no `code_verifier` at the token endpoint, it's going out as a plain code flow. Doesn't match the README or the PR description.

Should be a one line fix, nothing else needs touching. Other providers here use `$usesPKCE` (`src/Kick/Provider.php:19`, `src/Etsy/Provider.php:20`).

```suggestion
    protected $usesPKCE = true;
```

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+    /**
+     * Verify the JWT signature and decode it.
+     */
+    protected function verifyAndDecodeJWT(string $jwt, ?string $alg)
+    {
+        try {
+            JWT::$leeway = (int) ($this->getConfig('clock_skew') ?? 0);
+
+            $publicKey = $this->getConfig('jwt_public_key');
+            $configuredAlg = $this->getConfig('jwt_algorithm') ?: $alg ?: 'RS256';
```
atymic: **High**

`$alg` comes from `decodeJwtHeader($jwt)`, so it's straight off the unverified token header. When `jwt_algorithm` isn't set it ends up in `new Key($publicKey, $configuredAlg)` on the next line.

php-jwt's only algorithm check is `constantTimeEquals($key->getAlgorithm(), $header->alg)`, and here both sides come from the same attacker-supplied header, so it passes trivially. If someone's using `jwt_public_key` (a PEM public key, so not secret) you can forge `alg: HS256` and HMAC-sign it with the PEM string as the secret. `Key::getAlgorithm()` returns HS256, matches the header, `verify()` dispatches to `hash_hmac` with the PEM bytes as the key, and it verifies.

I ran this against v7.0.3 to be sure, replicating the two lines exactly, and a forged `sub=admin` token with no private key is accepted:

    legit RS256  => ACCEPTED as sub=alice
    FORGED HS256 => ACCEPTED as sub=admin

To be clear this isn't a php-jwt bug, the library is fine. The v6 algorithm-confusion fix works by binding key material to an algorithm in the `Key` object, and it assumes the caller pins that algorithm rather than reading it off the token (firebase/php-jwt#351). This is the RFC 8725 section 2.1 substitution attack reintroduced above the library. OIDC Core 3.1.3.7 step 7 lands in the same place, the alg should be RS256 or whatever the client registered as `id_token_signed_response_alg`, not what the token claims.

It also hits `verifyLogoutToken()` on line 681, which is the worse one since that endpoint is unauthenticated and the signature is the only thing standing in the way.

Pin it to `jwt_algorithm`, fall back to the OP's `id_token_signing_alg_values_supported`, then RS256, and reject `HS*` when the key material is an asymmetric public key.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+    {
+        return Str::random(40);
+    }
+
+    /**
+     * Determine if JWT signature verification is enabled.
+     */
+    protected function shouldVerifyJwt(): bool
+    {
+        return (bool) $this->getConfig('verify_jwt', $this->verifyJwt);
```
atymic: **High**

Flagging this as a hardening thing rather than a bug, since the spec does allow it. `verify_jwt` defaulting to false means `decodeJWT()` takes the else branch on line 310 and just base64 decodes the payload, so `sub`, `email`, `groups` and `role` are trusted as-is.

OIDC Core 3.1.3.7 step 6 explicitly permits this: "If the ID Token is received via direct communication between the Client and the Token Endpoint (which it is in this flow), the TLS server validation MAY be used to validate the issuer in place of checking the token signature." So the default is defensible on the happy path.

The catch is that the exemption is conditional on TLS, and nothing here checks it. `base_url` is used to build every endpoint and there's no scheme validation, so an `http://` base_url quietly turns the whole flow into "trust whatever came back" with no signature to fall back on. Given this is the generic provider people will copy from, defaulting to true and letting the odd OP that can't serve a JWKS opt out seems like the safer shape.

Separately, the TLS argument doesn't cover `verifyLogoutToken()` at all, since that token arrives unsolicited from the internet. That path already forces verification, which is right, it just means the two paths have different trust models and that's worth a comment in the code.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopeSeparator = ' ';
+
+    /**
+     * Indicates if the nonce should be utilized.
+     */
+    protected bool $usesNonce = true;
```
atymic, line 53: **High**

`stateless()` can't work here. `$usesNonce` is hardcoded true, it's not in `additionalConfigKeys()`, and there's no setter, so there's no way to turn it off.

In stateless mode `redirect()` skips the state on line 84 but still writes the nonce to the session on line 88, and `getCurrentNonce()` reads it back from the session on the callback. So for an API/SPA flow doing `->stateless()->user()` there's no session to carry it, `getCurrentNonce()` returns null, and `isInvalidNonce()` needs `$nonce === $this->getCurrentNonce()` with `is_string()` on top, which can never be true. Every login throws `JWT: Contains an invalid nonce.`

If there's genuinely no session bound it'll blow up on `$this->request->session()` instead. Either way it needs a toggle, and the nonce should be skipped when `isStateless()`.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+            throw new InvalidArgumentException('Callback: invalid state.', 401);
+        }
+
+        if (! $this->request->filled('code')) {
+            throw new InvalidArgumentException('Callback: missing authorization code.', 401);
+        }
+
+        $tokenResponse = $this->getAccessTokenResponse($this->request->input('code'));
+
+        $payload = $this->decodeJWT($tokenResponse['id_token'], $tokenResponse['access_token'] ?? null);
```
atymic: **Medium**

Lines 283, 286, 293, 297 and 299 index `$tokenResponse` directly.

Line 299 is the one most likely to bite: `expires_in` is optional per RFC 6749 4.2.2 and a fair few IdPs leave it out, so a login that otherwise worked fine dies on an undefined key. Line 286 has the same issue when only an `id_token` comes back.

There's also the IdP that returns 200 with an error body rather than a 4xx (Guzzle catches the 4xx case for you, not this one). That hits line 283 with no `id_token`, which is a warning followed by a `TypeError` since the param is typed `string`.

Both Socialite's and the Manager's `user()` use `Arr::get()` for this. Probably worth matching, plus an explicit check for `id_token` so the error says something useful.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+     * Validate exp/nbf/iat with configurable leeway. Runs in both verified
+     * and unverified paths so a stale id_token is never accepted even when
+     * signature verification is disabled.
+     */
+    protected function validateTimeClaims($payload): void
+    {
+        $now = time();
+        $leeway = (int) ($this->getConfig('clock_skew') ?? 0);
+
+        if (isset($payload->exp) && $now - $leeway >= (int) $payload->exp) {
```
atymic: **Medium**

All three checks are behind `isset()`, but `exp` is required, not optional. OIDC Core 3.1.3.7 step 9 says the current time MUST be before `exp`, and Back-Channel Logout 2.4 requires both `exp` and `iat` on a logout token.

php-jwt won't save you either, its `decode()` gates `exp`/`nbf`/`iat` on `isset()` the same way. So a token that just omits `exp` passes both the verified and unverified paths and is good forever. With `verify_jwt` defaulting to false, a captured id_token with no `exp` is replayable indefinitely.

Should require `exp` outright, and `iat` too in `verifyLogoutToken()`.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+            throw new InvalidArgumentException('JWT: Invalid issuer.', 401);
+        }
+
+        $aud = $payload->aud ?? null;
+        $audList = is_array($aud) ? $aud : [$aud];
+        if (! in_array($this->clientId, $audList, true)) {
+            throw new InvalidArgumentException('JWT: Invalid audience.', 401);
+        }
+
+        if (is_array($aud) && count($aud) > 1 && ($payload->azp ?? null) !== $this->clientId) {
```
atymic: **Medium**

The `count($aud) > 1` gate isn't what the spec says. 3.1.3.7 step 5 is "when an `azp` claim is present, the Client SHOULD verify that its `client_id` is the Claim Value" — no audience count involved. The multi-audience rule in section 2 is an issuance requirement on the OP, not a validation rule for us.

So a token with `aud: "our-client-id"` and `azp: "some-other-client"` (authorized for a different client, we're just named as an audience) short-circuits before the azp check and gets accepted.

```suggestion
        if (isset($payload->azp) && $payload->azp !== $this->clientId) {
            throw new InvalidArgumentException('JWT: Invalid authorized party (azp).', 401);
        }
```

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+
+        return new RedirectResponse($this->getAuthUrl($state));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getScopes(): array
+    {
+        if ($this->getConfig('scopes')) {
```
atymic: **Medium**

`array_merge` with no dedupe, so config scopes of `openid email offline_access` sends `scope=openid email profile openid email offline_access`. Azure AD B2C rejects repeated scope values, and it's malformed either way.

The other half of it is you can't narrow the set. If an OP doesn't support `profile` it'll reject every authorization request and there's no config lever to drop it.

`array_values(array_unique(...))` at minimum. Might also be worth having an explicit `scopes` config replace the defaults rather than extend them, keeping `openid` since that one's mandatory.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+                'token'           => $token,
+                'token_type_hint' => $tokenTypeHint,
+                'client_id'       => $this->clientId,
+                'client_secret'   => $this->clientSecret,
+            ])
+        );
+
+        // RFC 7009: a successful response is 200, regardless of whether the
+        // token was valid. Some IdPs return 204.
+        return in_array($response->getStatusCode(), [200, 204], true);
```
atymic, line 1176: **Medium**

`revoke()` can't return false. Guzzle's `http_errors` is on by default so any 4xx/5xx throws before this line, which means the only way to reach the `in_array()` is a 2xx and it's always true.

The README example on line 892 calls `revoke($refresh, 'refresh_token')` with no try/catch, so an IdP that returns 400 for an already-revoked token throws an uncaught `ClientException` out of the logout controller. User gets a 500 instead of being logged out.

Either set `RequestOptions::HTTP_ERRORS => false` and actually branch on the status, or drop the bool return and document that it throws.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+     * client_secret_basic).
+     */
+    protected function tokenRequestOptions(array $fields): array
+    {
+        $method = $this->resolveTokenAuthMethod();
+
+        $options = [RequestOptions::HEADERS => ['Accept' => 'application/json']];
+
+        if ($method === 'client_secret_basic') {
+            $options[RequestOptions::AUTH] = [$this->clientId, $this->clientSecret];
```
atymic: **Low**

Guzzle's `auth` option does a plain `base64_encode("$user:$pass")` (Client.php around line 408). RFC 6749 2.3.1 wants the client id and secret each form-urlencoded before they're joined and base64'd.

Doesn't matter for alphanumeric secrets, but breaks on anything with `+`, `/`, `=`, a space, or non-ASCII, and base64-derived secrets with `+` or `/` in them are pretty common. Comes back as an opaque `invalid_client` which is horrible to trace.

`resolveTokenAuthMethod()` on line 573 prefers basic whenever the OP advertises it, so this is the default path for most IdPs. Building the header by hand would do it:

```php
'Authorization' => 'Basic '.base64_encode(urlencode($this->clientId).':'.urlencode($this->clientSecret))
```

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+        }
+
+        $params = array_filter(array_merge([
+            'id_token_hint'            => $idToken,
+            'client_id'                => $this->clientId,
+            'post_logout_redirect_uri' => $postLogoutRedirectUri ?? $this->getConfig('post_logout_redirect_uri'),
+            'state'                    => $state,
+        ], $extra), fn ($v) => $v !== null && $v !== '');
+
+        return new RedirectResponse($config['end_session_endpoint'].'?'.http_build_query($params));
```
atymic: **Low**

The `'?'` is hardcoded. Some multi-tenant deployments advertise an endpoint that already has a query on it, e.g. `https://idp.example.com/logout?realm=prod`, and you'd end up with `.../logout?realm=prod?id_token_hint=...`. Everything after the second `?` gets parsed into the `realm` value, so all the logout params silently vanish.

`.(str_contains($endpoint, '?') ? '&' : '?').` would sort it.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+                throw new InvalidArgumentException('JWT: User has no email.', 401);
+            }
+        }
+
+        $raw = (array) $payload;
+        $raw['id_token'] = $tokenResponse['id_token'];
+
+        $this->user = $this->mapUserToObject($raw);
+
+        return $this->user->setToken($tokenResponse['access_token'])
```
atymic: **Low**

This override reimplements `user()` from scratch and loses two things the Manager's base sets: `setAccessTokenResponseBody($this->credentialsResponseBody)` and `setApprovedScopes($this->parseApprovedScopes($response))`. `$this->credentialsResponseBody` never gets assigned either.

So anything written against the Manager's `User` API, `$user->accessTokenResponseBody['id_token']` or checking `$user->approvedScopes` to see whether `offline_access` was actually granted, comes back null for this provider and no other.

The README works around it by putting `id_token` into the raw array on line 293, which leaves the same data in two places, one of them non-standard.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+        $expected = $this->base64UrlEncode(substr($digest, 0, intdiv(strlen($digest), 2)));
+
+        if (! hash_equals($expected, $atHash)) {
+            throw new InvalidArgumentException('JWT: at_hash mismatch.', 401);
+        }
+    }
+
+    private function base64UrlDecode(string $data): string
+    {
+        return base64_decode(str_pad(strtr($data, '-_', '+/'), strlen($data) % 4, '=', STR_PAD_RIGHT));
```
atymic: **Low**

`str_pad`'s second arg is the target total length, not the number of pad chars. `strlen($data) % 4` is 0-3 and always `<= strlen($data)`, so this returns the string untouched and no `=` ever gets added.

Nothing actually breaks today because PHP's non-strict `base64_decode()` is fine with missing padding (I fuzzed it across the length classes, every round trip matched). But it reads like it's doing something, and it'd break the moment anyone switched to `base64_decode($x, true)`.

```suggestion
        return base64_decode(str_pad(strtr($data, '-_', '+/'), intdiv(strlen($data) + 3, 4) * 4, '=', STR_PAD_RIGHT));
```

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+     */
+    public function logout(?string $idToken = null, ?string $postLogoutRedirectUri = null, array $extra = []): RedirectResponse
+    {
+        $config = $this->getOpenIdConfig();
+
+        if (empty($config['end_session_endpoint'])) {
+            throw new InvalidArgumentException('Provider does not advertise an end_session_endpoint.');
+        }
+
+        $state = Str::random(40);
```
atymic: **Low**

`logout_state` gets generated, stored and sent as the `state` param, but nothing validates it when the IdP redirects back to `post_logout_redirect_uri`, and the README doesn't tell the caller to check it either.

As it stands it looks like CSRF protection on the logout round-trip without being any. Either add a helper (`validateLogoutState(Request $request): bool` doing a `hash_equals`) and document it in the RP-Initiated Logout section, or drop the state.

src/OpenIDConnect/Provider.php
```diff
@@ -0,0 +1,715 @@
...
+use GuzzleHttp\RequestOptions;
+use Illuminate\Http\RedirectResponse;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Str;
+use InvalidArgumentException;
+use JsonException;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
```
atymic, line 23: **Low**

`vendor/bin/pint --test src/OpenIDConnect/` fails on this file. Pending fixers: `fully_qualified_strict_types`, `is_null`, `unary_operator_spaces`, `yoda_style`, `not_operator_with_successor_space`, `ordered_imports`, `phpdoc_align`.

`fully_qualified_strict_types` is the only one that's more than style, it's flagging the inline `new \GuzzleHttp\Client(...)` on line 195 which should be a `use GuzzleHttp\Client;` next to the other Guzzle imports. `composer lint` clears the lot.

atymic (changes requested):
Had a careful read through, mostly against the OIDC Core / Discovery / Back-Channel Logout specs and the vendored firebase/php-jwt source. Really like where this is going, a generic OIDC provider is a genuinely useful addition and the spec coverage here is more thorough than most of what's in this repo.

15 comments inline. Four I'd want sorted before merge:

- **PKCE isn't on.** `$pkceEnabled` isn't the property the base class reads, so the flag never flips and it's going out as a plain code flow. One line fix.
- **Algorithm confusion in `verifyAndDecodeJWT()`.** The signing alg is taken from the unverified token header. I wrote a PoC against v7.0.3 and it forges an admin token with no private key. This one's the reason for the request-changes.
- **`stateless()` can't work**, the nonce is hardcoded on and always read from the session.
- **`exp` is only checked when present**, so a token that omits it never expires.

The rest are medium/low: unguarded array access on the token response, the azp gate, scope duplication, `revoke()` never returning false, and some smaller stuff. Comments 1, 7 and 13 have suggestions attached.

Happy to be wrong on any of these, and glad to help with the alg pinning if useful.

src/OpenIDConnect/Provider.php
adrum: Fixed, and reproduced against php-jwt v7.1.0 first — a forged sub=admin token was accepted.

One addition: pinning alone isn't sufficient. If an OP advertises both RS256 and HS256 in id_token_signing_alg_values_supported, HS256 is legitimately in the allow list and the forgery walks straight back in. So there are two further guards: HS* is rejected when the key material is asymmetric (PEM/cert), and rejected outright on the JWKS path, since a JWKS never publishes an HMAC secret. jwt_algorithm now accepts a list, and none is always dropped.

src/OpenIDConnect/Provider.php
adrum: Fixed — defaults to true, and getBaseUrl() now rejects non-https issuers (loopback exempt for local dev). Trust models documented on both paths as suggested.

Worth flagging: flipping the default naively would have made opt-out impossible. ConfigTrait::getConfig() uses empty(), so verify_jwt => false is indistinguishable from unset and falls through to the default. shouldVerifyJwt() reads the raw config array instead.

src/OpenIDConnect/Provider.php
adrum: Fixed — Arr::get() throughout, matching Socialite/Manager, plus explicit checks for an error body and a missing id_token. The userinfo fallback is skipped when there's no access token to make it with.

src/OpenIDConnect/Provider.php
adrum: Fixed — exp required (missing or non-numeric), iat required in verifyLogoutToken().

Note your last line is now stale in your favor: verify_jwt defaults to true after item 3, but that didn't reduce the bug at all — php-jwt gates its own exp check on isset(), so the verified path was equally affected. iat left optional for id_tokens, as you scoped it, though OIDC Core §2 lists it as REQUIRED if you want that too.

src/OpenIDConnect/Provider.php
adrum: Fixed — deduped, and a configured list now replaces the defaults (openid always sent). Parsing also accepts arrays and comma separation; explode(' ', ...) meant 'email, offline_access' produced a scope literally named "email,".

One interaction: replace-semantics would have silently dropped fluent ->scopes([...]) additions. Diffing against a DEFAULT_SCOPES constant instead of the live list preserves them. This is a behavior change — scopes no longer extends the defaults.

src/OpenIDConnect/Provider.php
adrum: Fixed — http_errors => false and branch on status, so a 400 for an already-revoked token returns false instead of throwing out of the logout controller.

Transport failures still throw: an unreachable IdP is a different thing from a rejected revocation, and silently returning false would hide an outage. README example wraps the call accordingly.

src/OpenIDConnect/Provider.php
adrum: Fixed — header built by hand with urlencode() on each half. Confirmed Guzzle's raw base64_encode("$id:$secret") at Client.php:1193 (line differs from your ~408, version drift). urlencode rather than rawurlencode is right here: RFC 6749 2.3.1 specifies the application/x-www-form-urlencoded algorithm, where space is +.

src/OpenIDConnect/Provider.php
adrum: Fixed. Also applied to authorization_endpoint — Socialite's buildAuthUrlFromBase() hardcodes ? identically, so the same multi-tenant deployment had the same bug on login, which is the worse of the two. Both route through a shared appendQuery().

src/OpenIDConnect/Provider.php
adrum: Fixed — credentialsResponseBody assigned, setAccessTokenResponseBody() and setApprovedScopes() restored, using the base parse* helpers.

Dropped the id_token raw injection entirely rather than keeping both. Per OIDC Core 3.1.3.3 the id_token is a member of the token response, and per §5.1/§2 the raw user is the claim set — a token isn't a claim. README updated to read from accessTokenResponseBody.

src/OpenIDConnect/Provider.php
adrum: Fixed — added validateLogoutState(Request $request): bool using hash_equals, documented in the RP-Initiated Logout section. It pulls the stored value, so a state is good for one round trip and a replay fails.

Also stopped emitting a state at all when there's no session to store it in — previously one was generated and sent regardless, which is the same appearance-of-protection in the one case where protection is impossible.

src/OpenIDConnect/Provider.php
adrum: Should be good on pint now.

src/OpenIDConnect/Provider.php
adrum: Fixed — agree the count gate conflates two separate rules.

Your snippet as written would close step 5 but drop step 4: the current condition enforces "multiple audiences require an azp" by accident (absent azp compares null against the client id and throws), and isset($payload->azp) && ... stops rejecting that. Verified both ways against tests. Shipped both clauses explicitly with distinct messages.

src/OpenIDConnect/Provider.php
adrum: Fixed — usesNonce() returns false when isStateless(), plus a use_nonce config key and withoutNonce().

Fixing the nonce alone only moved the crash: Socialite's PKCE is session-bound too (code_verifier written at redirect, read in getCodeChallenge()/getTokenFields()), so a session-less flow still died with the same opaque error. redirect() now reports the real requirement, and withoutPKCE() was added since Socialite ships enablePKCE() with no counterpart.

adrum:
@atymic I have a bunch of automated tests ready to go for this PR, but noticed the repo has no test suite — composer.json declares test/test:ci scripts but phpunit isn't in require-dev, and the Test workflow's only job runs parallel-lint and composer validate. Are you open to a follow-up PR introducing automated tests for this provider, wired into GitHub Actions?

Two changes in this PR that weren't in the review, both in user():
- UserInfo sub now verified (OIDC Core 5.3.2). The fallback replaced the payload with whatever userinfo returned without checking its sub against the id_token's, so a response carrying a different sub would silently substitute another user's claims. Now verified, and merged over the id_token rather than replacing it, so groups/role survive the fallback.
- email is no longer required to log in. user() threw whenever no email was obtainable, but email depends on the email scope being granted and OIDC Core §2 makes sub the only guaranteed identifier — a conforming deployment that doesn't release email couldn't authenticate at all. Identity now keys on sub; require_email restores the old behaviour. Flagging explicitly since it changes login behaviour and isn't a fix you asked for.

Three things I looked at and deliberately didn't change:
- iat on id_tokens — left optional per your scoping, though OIDC Core §2 lists it as REQUIRED alongside exp. I kept to what you cited rather than tightening unprompted, since unlike the email change it would reject tokens that currently work. Happy to tighten it.
- Stateless + PKCE — works only when a session store happens to be bound. Proper support means letting the caller supply and return the verifier, which is upstream Socialite's concern more than this provider's.
- appendQuery() and fragments — uses str_contains($url, '?'), so a URL with a fragment would misplace the query. No discovery document should advertise one on these endpoints.

atymic (commented):
> Are you open to a follow-up PR introducing automated tests for this provider, wired into GitHub Actions?

Yes, actually that would be amazing, been meaning to add support for it to this repo.
It's a little bit complicated since all the providers are split/etc, I will do some research on the best way (I guess each provider with tests needs a composer install)

atymic:
@adrum 
See https://github.com/SocialiteProviders/Providers/pull/1472 if you add tests they will run :)

kohenkatz:
A fix for this may be something that should go in another PR after this one, but I am testing this and I ran into the following issue:

When using Microsoft Entra OIDC in "multi-tenant" mode, the `issues` in the OpenID configuration is always `https://login.microsoftonline.com/{tenantid}/v2.0`, but the returned token puts in the actual tenant ID, e.g. `https://login.microsoftonline.com/aaaa1111-bbbb-2222-cccc-dddd3333eeee/v2.0`. This makes the token validation fail with `JWT: Invalid issuer`.

I have already dealt with the same problem on Android and iOS (using Okta's OIDC library) by creating a custom IdTokenValidator implementation that replaces `{tenantid}` with the actual tenant ID (from the `tid` field), and then continuing the validation as usual.

It would be nice to have a way to override the validation logic to add in conditions like this. (I have no idea if anyone but Microsoft has this issue, but Microsoft is certainly a large enough authentication source to try to make this work.

Many places in Laravel already have the concept of plugging in a function that resolves some value or performs some action, but I'm not sure whether to suggest that the _entire_ `validateIdTokenClaims` method should be pluggable or whether just having the `issuer` validation be pluggable is enough. On one hand, it would be nice to not have to copy over the entire function body to make one change, but on the other hand, what other validation changes might other providers require (other than for `issuer`) that could make it make sense to allow overriding the whole thing?

atymic:
TBH i'm almost thinking at this point maybe this should be a seperate, but compatible with, socialite providers package.
Ill have a look today.

atymic:
@adrum have a look at:
https://github.com/SocialiteProviders/openid-connect/pull/1

adrum:
@atymic I forgot to push last night. I pushed here for continuity. Do you want me to close this?

atymic:
> @atymic I forgot to push last night. I pushed here for continuity. Do you want me to close this?
Thanks for your work! I will use the magic of claude to check all your tests apply to the new package, it should be close as it's mostly the work here.
I'lll add you as the provider author given you wrote most the code also :)
A review of the new repo would be great if you have a chance.

atymic:
@adrum 
I've cross checked and added some extra tests to the PR based on your tests, I believe all the cases are fully covered.

atymic:
Thanks all 🙏 

Sorry for the PR mess. Merged in https://github.com/SocialiteProviders/Providers/pull/1475
Any follow up PR's are welcome, I believe i've addressed any feedback though :)

PR #1452, opened 2026-06-15
feat(saml2): support manual IdP Single Logout via "slo" config key

mbardelmeijer:
## Why

Single Logout currently only works when the Identity Provider is configured through `metadata`. When the IdP is configured **manually** with the `acs` key, the descriptor is built by lightsaml's `SimpleEntityDescriptorBuilder`, which only emits a `SingleSignOnService` — it has no concept of a `SingleLogoutService`.

As a result, `logoutResponse()` cannot resolve the IdP's logout endpoint (`getFirstSingleLogoutService()` returns nothing), so IdP-initiated SLO is impossible for manually-configured providers unless you supply full IdP metadata XML.

## What

Adds an optional `slo` config key holding the IdP's Single Logout URL. When present, `getIdentityProviderEntityDescriptorManually()` advertises a `SingleLogoutService` on the IdP descriptor for both HTTP-Redirect and HTTP-POST bindings, so the existing native `logoutResponse()` flow can resolve it.

This mirrors how `acs` already provides the IdP's SSO location for the manual path (`acs` ↔ SSO, `slo` ↔ SLO).

```php
'saml2' => [
  'acs' => 'https://idp.co/auth/sso',
  'slo' => 'https://idp.co/auth/slo', // new
  'entityid' => 'http://saml.to/trust',
  'certificate' => '...',
],
```

## Notes

- **Fully backward compatible** — behaviour is unchanged when `slo` is omitted (no `SingleLogoutService` is added).
- No new imports required: `SingleLogoutService` and `SamlConstants` are already imported in `Provider.php`.
- Both bindings are advertised so `logoutResponse()` resolves the endpoint regardless of the configured `idp_binding_method`.
- README updated to document `slo` in the Single Logout section.

### DIFF
--- a/src/Saml2/Provider.php
+++ b/src/Saml2/Provider.php
@@ -149,6 +149,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
             'metadata',
             'ttl',
             'acs',
+            'slo',
             'entityid',
             'certificate',
             'sp_acs',
@@ -270,6 +271,7 @@ class Provider extends AbstractProvider implements SocialiteProvider
     protected function getIdentityProviderEntityDescriptorManually(): EntityDescriptor
     {
         $acs = $this->getConfig('acs');
+        $slo = $this->getConfig('slo');
         $entityId = $this->getConfig('entityid');
         $certificate = $this->getConfig('certificate');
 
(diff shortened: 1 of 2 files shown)


atymic:
@27pchrisl love some eyes here :)

27pchrisl:
Looks good to me @atymic, thanks for the PR @mbardelmeijer !

atymic (approved):

PR #1453, opened 2026-06-16
SAML2: support service provider certificate overlap rotation

mbardelmeijer:
## Why this is needed

When a Laravel application acts as a SAML service provider with signed `AuthnRequest` messages, identity providers must trust the SP's signing certificate ("request signer certificate"). Many IdPs — especially enterprise and government setups — register that certificate once and pin it, rather than continuously refreshing SP metadata.

Today this package only supports a single `sp_certificate` / `sp_private_key` pair. Replacing those values is therefore a hard cutover: metadata and outbound signatures switch to the new key immediately, while IdPs that still trust only the old certificate reject login requests until an administrator manually updates their configuration.

That is a problem in normal operations, not just at initial setup:

- SP signing certificates expire and must be rotated on a schedule (often every 1–3 years).
- Compliance requirements may mandate shorter lifetimes or replacement of non-conforming certificates.
- Operators need a grace period so customers can update pinned certificates without an SSO outage.

The package already handles the **inbound** side of this well: when an IdP publishes multiple signing certificates in metadata, `validateSignature()` accepts any of them. There is no equivalent for the **outbound** SP signing certificate, which is what this PR adds.

## What this changes

Adds optional overlap credentials via three new config keys:

- `sp_previous_certificate`
- `sp_previous_private_key`
- `sp_previous_private_key_passphrase`

During a rotation overlap period, operators deploy the new active `sp_*` credentials and keep the replaced pair in `sp_previous_*` for a grace window (typically 30–90 days).

**Metadata publication:** SP metadata includes `KeyDescriptor` entries for both the active and previous certificates (signing and encryption).

**Outbound signing:** `AuthnRequest`, `LogoutRequest`, and `LogoutResponse` messages are signed with the active key only. The entity descriptor signature also uses the active key.

**Inbound decryption:** If assertions are encrypted, decryption is attempted with the active private key first, then the previous private key.

**Application API:** Adds `getServiceProviderCertificates()` so applications can expose all currently published SP certificates (for example on a `/certificate` endpoint) without parsing metadata XML.

`credential()` is preserved as an alias to the active signing credential, so existing behaviour is unchanged when `sp_previous_*` is not configured.

## Recommended rotation workflow

1. Generate and deploy a new active `sp_certificate` / `sp_private_key`.
2. Move the replaced certificate and private key to `sp_previous_*`.
3. Notify IdP administrators to refresh SP metadata or update their pinned request-signer certificate.
4. After the grace period, remove `sp_previous_*`.

Documentation for this workflow is added to the SAML2 README.


mbardelmeijer:
@27pchrisl if you're able to take a look at this, that would be great! 🙏

atymic:
cc @27pchrisl

src/Saml2/Provider.php
```diff
@@ -684,6 +699,14 @@ public function getServiceProviderMetadata(): Response
             ->setContent($serializationContext->getDocument()->saveXML());
     }
 
+    /**
+     * @return X509Certificate[]
+     */
+    public function getServiceProviderCertificates(): array
```
27pchrisl: What is the purpose of these public functions? If there's a need for the visibility to change for external code to make use of them, the visibility could just be changed on `metadataCertificates` without the extra getter

src/Saml2/Provider.php
```diff
@@ -657,20 +665,27 @@ protected function receive(): void
...
 
         if ($reader === null) {
             return;
         }
 
-        $assertion = $reader->decryptAssertion($credential->getPrivateKey(), new DeserializationContext);
-        $this->messageContext->asResponse()->addAssertion($assertion);
+        $lastException = null;
+
+        foreach ($this->decryptionCredentials() as $credential) {
```
27pchrisl: `EncryptedAssertionReader` has a `decryptMultiAssertion` to handle key rollover - there's no need for the foreach here because you can pass all the possible credentials into that function. It also throws the right exceptions so we don't need our own exception handling here.

src/Saml2/Provider.php
```diff
@@ -701,15 +724,87 @@ protected function signature(X509Credential $credential): SignatureWriter
 
     protected function credential(): ?X509Credential
     {
-        if (! $this->getConfig('sp_certificate') || ! $this->getConfig('sp_private_key')) {
+        return $this->signingCredential();
```
27pchrisl: Similar to above, there's no need for a function here that just calls another function.

src/Saml2/Provider.php
mbardelmeijer: Good call; was using this public method for a test only. Have reverted the change.

src/Saml2/Provider.php
mbardelmeijer: Ah, good to know! Have modified it to use that instead.

mbardelmeijer:
@27pchrisl requested changes have been made :)

src/Saml2/Provider.php
```diff
@@ -699,17 +701,84 @@ protected function signature(X509Credential $credential): SignatureWriter
         );
     }
 
-    protected function credential(): ?X509Credential
+    protected function signingCredential(): ?X509Credential
```
27pchrisl: @mbardelmeijer one more here - could you change the function name back to ‘credential’ ? Then I think this is good to merge

mbardelmeijer:
@27pchrisl change has been made, thanks!

27pchrisl:
@atymic This looks good to merge now, thanks for the contribution @mbardelmeijer !

atymic:
Thanks team!

PR #1455, opened 2026-06-26
Microsoft provider: add public getClaims and getGroups function

Jimbolino:
I saw that the current microsoft provider already fetches some claims from the id_token (roles and tenant-id)
For a customer of ours we were looking at using the "groups" claim.
Because there are [many more claims](https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims-reference), i think the class would benefit from a public function to access them.

### DIFF
--- a/src/Microsoft/Provider.php
+++ b/src/Microsoft/Provider.php
@@ -24,6 +24,8 @@ class Provider extends AbstractProvider
 
     private ?array $jwtKeys = null;
 
+    private ?\stdClass $claims = null;
+
     /**
      * The tenant id associated with personal Microsoft accounts (services like Xbox, Teams for Life, or Outlook).
      * Note: only reported in JWT ID_TOKENs and not in call's to Graph Organization endpoint.
@@ -165,6 +167,7 @@ class Provider extends AbstractProvider
         }
 
         $formattedResponse['roles'] = $this->getRoles();
+        $formattedResponse['groups'] = $this->getGroups();
 
         return $formattedResponse;
     }
@@ -205,6 +208,7 @@ class Provider extends AbstractProvider
             'userPrincipalName' => Arr::get($user, 'userPrincipalName'),
             'employeeId'        => Arr::get($user, 'employeeId'),
             'roles'             => Arr::get($user, 'roles'),
+            'groups'            => Arr::get($user, 'groups'),
 
             'tenant' => Arr::get($user, 'tenant'),
         ]);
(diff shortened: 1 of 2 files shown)


src/Microsoft/Provider.php
```diff
@@ -233,6 +233,28 @@ public static function additionalConfigKeys(): array
...
+     * Get all claims from the ID_TOKEN
+     */
+    public function getClaims()
+    {
+        if ($idToken = $this->parseIdToken($this->credentialsResponseBody)) {
+
+            return $this->validate($idToken);
+        }
+
+        return [];
```
atymic: `getClaims()` returns two shapes, `validate()` gives a `stdClass` and the fallback gives `[]`.

Works today because `[]?->groups` is null on PHP 8, but for a public method it's a bad contract. `$provider->getClaims()->email` fatals on the no-token path with no way to tell why. Return null and type it, which also makes the `?->` in the two callers meaningful instead of accidental.

```suggestion
    /**
     * Get all claims from the ID_TOKEN.
     */
    public function getClaims(): ?\stdClass
    {
        if ($idToken = $this->parseIdToken($this->credentialsResponseBody)) {
            return $this->validate($idToken);
        }

        return null;
    }
```

src/Microsoft/Provider.php
```diff
@@ -233,6 +233,28 @@ public static function additionalConfigKeys(): array
...
+     * Get the optional groups from the ID_TOKEN.
+     * https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims-reference
+     */
+    public function getGroups()
+    {
+        return $this->getClaims()?->groups ?? [];
+    }
+
     /**
      * Get user's roles from the ID_TOKEN.
```
atymic, line 268: Both new methods are untyped, everything else public here is typed (`getRoles(): array`, `isConsumerTenant(): bool`).

```suggestion
    public function getGroups(): array
```

src/Microsoft/Provider.php
```diff
@@ -257,14 +274,7 @@ public function getRoles(): array
...
     {
-        if ($idToken = $this->parseIdToken($this->credentialsResponseBody)) {
-
-            $claims = $this->validate($idToken);
-
-            return ($claims?->tid ?? '') === self::MS_ENTRA_CONSUMER_TENANT_ID;
-        }
-
-        return false;
+        return ($this->getClaims()?->tid ?? '') === self::MS_ENTRA_CONSUMER_TENANT_ID;
```
atymic, line 286: Heads up rather than an objection: no `id_token` used to mean an early `return false` with no call to `validate()`. Now it always runs. Same result (`'' !== '9188…'`), but it's a full JWT validation on a path that was a no-op. See the memoisation comment.

src/Microsoft/Provider.php
```diff
@@ -233,6 +233,28 @@ public static function additionalConfigKeys(): array
         ];
     }
 
+    /**
+     * Get all claims from the ID_TOKEN
+     */
+    public function getClaims()
```
atymic: This one I'd call a requirement.

`validate()` does a JWKS parse, `JWT::decode()` signature check, plus iss/aud/exp. `getUserByToken()` already hits `isConsumerTenant()` (L148) and `getRoles()` (L167), so that's two full validations per login. This PR doesn't make it worse, but making `getClaims()` public means userland can call `getClaims()`/`getGroups()`/`getRoles()` and revalidate each time.

The class already memoises `$openIdConfiguration` and `$jwtKeys`, so caching the payload fits:

```php
private ?\stdClass $claims = null;

public function getClaims(): ?\stdClass
{
    if ($this->claims !== null) {
        return $this->claims;
    }

    if ($idToken = $this->parseIdToken($this->credentialsResponseBody)) {
        return $this->claims = $this->validate($idToken);
    }

    return null;
}
```

src/Microsoft/Provider.php
```diff
@@ -233,6 +233,28 @@ public static function additionalConfigKeys(): array
...
+
+    /**
+     * Get the optional groups from the ID_TOKEN.
+     * https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims-reference
+     */
+    public function getGroups()
+    {
+        return $this->getClaims()?->groups ?? [];
+    }
+
```
atymic, line 266: Link's right for `groups`. While you're here, the `getRoles()` docblock below points at `optional-claims#configure-groups-optional-claims`, the groups anchor on the roles method. Pre-existing, but easy to fix since you're touching both.

src/Microsoft/Provider.php
```diff
@@ -233,6 +233,28 @@ public static function additionalConfigKeys(): array
         ];
     }
 
+    /**
```
atymic, line 240: Can you add these to `src/Microsoft/README.md`? `getRoles()` is documented at L71.

`getGroups()` needs the caveat that the claim isn't emitted by default, it has to be turned on in the Entra app registration, and when a user is in too many groups (>200 for JWT) Entra drops it and sends `_claim_names`/`_claim_sources` pointing at Graph instead. Without that you'll get "getGroups() returns []" issues.

atymic (changes requested):
Good cleanup, the duplication between `getRoles()` and `isConsumerTenant()` was worth removing, and this fixes the undefined `$claims` path in `getRoles()` along the way.

Few things before it goes in, mostly because `getClaims()` is now public API. Comments inline.

Also, not blocking: `roles` is on the user object via `mapUserToObject()` (L207). Should `groups` be too, so `$user->groups` works?

PR #1458, opened 2026-07-12
feat: add Epic Games provider

SGzuis:
### What does this PR do?
This PR introduces a new OAuth2 provider for **Epic Games**, allowing developers to easily integrate Epic
Games authentication into their Laravel applications using Socialite.

### Implementation Details:
- **Provider ID:** `EPIC_GAMES`
- **Auth URL:** `https://www.epicgames.com/id/authorize`
- **Token URL:** `https://api.epicgames.dev/epic/oauth/v2/token`
- **User Info URL:** `https://api.epicgames.dev/epic/oauth/v2/userInfo`
- Handled the Epic Games requirement to send credentials via `Basic Auth` in the token exchange request
header instead of passing them in the POST body.
- Mapped standard User fields (`id`, `nickname`, `name`, `email`) corresponding to the Epic Games user info
endpoint schema.

### Checklist:
- [x] Provider class created.
- [x] Listener (`EpicGamesExtendSocialite`) created.
- [x] `composer.json` configured appropriately.
- [x] `README.md` created with usage and installation instructions.

### DIFF
--- a/src/EpicGames/Provider.php
+++ b/src/EpicGames/Provider.php
@@ -4,8 +4,8 @@ namespace SocialiteProviders\EpicGames;
 
 use GuzzleHttp\RequestOptions;
 use Illuminate\Support\Arr;
-use Laravel\Socialite\Two\User;
 use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
 
 class Provider extends AbstractProvider
 {
@@ -92,7 +92,6 @@ class Provider extends AbstractProvider
             'nickname' => Arr::get($user, 'preferred_username'),
             'name' => Arr::get($user, 'name', Arr::get($user, 'preferred_username')),
             'email' => Arr::get($user, 'email'),
-            'avatar' => null,
         ]);
     }
 }
--- a/src/EpicGames/Provider.php
+++ b/src/EpicGames/Provider.php
@@ -1,0 +1,98 @@
+<?php
+
+namespace SocialiteProviders\EpicGames;
+
+use GuzzleHttp\RequestOptions;
+use Illuminate\Support\Arr;
+use Laravel\Socialite\Two\User;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+
+class Provider extends AbstractProvider
+{
+    public const IDENTIFIER = 'EPIC_GAMES';
+
+    /**
+     * The scopes being requested.
+     *
+     * @var array
+     */
+    protected $scopes = ['basic_profile'];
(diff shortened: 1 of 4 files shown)


src/EpicGames/Provider.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
+namespace SocialiteProviders\EpicGames;
+
+use GuzzleHttp\RequestOptions;
+use Illuminate\Support\Arr;
```
atymic, line 6: Other providers in this repo import `SocialiteProviders\Manager\OAuth2\User` (it extends the Socialite `User`). Mind switching for consistency?

src/EpicGames/Provider.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    protected function mapUserToObject(array $user): User
+    {
+        return (new User())->setRaw($user)->map([
+            'id' => Arr::get($user, 'sub'),
+            'nickname' => Arr::get($user, 'preferred_username'),
+            'name' => Arr::get($user, 'name', Arr::get($user, 'preferred_username')),
+            'email' => Arr::get($user, 'email'),
+            'avatar' => null,
+        ]);
```
atymic, line 95: Minor: Epic's userinfo has no avatar and the README doesn't list it, so this line can just be dropped rather than hardcoded to `null`.

src/EpicGames/README.md
```diff
@@ -0,0 +1,49 @@
...
+```bash
+composer require socialiteproviders/epic-games
+```
+
+## Installation & Basic Usage
+
+Please see the [Base Installation Guide](https://socialiteproviders.com/usage/), then follow the provider specific instructions below.
+
+### Add configuration to `config/services.php`
+
```
atymic, line 12: The driver registers as `epic-games`, and Socialite resolves the config key from the driver name verbatim — so `driver('epic-games')` reads the `epic-games` key, not `epic_games`. I think this should be `'epic-games'`; can you confirm? As written the config lookup would silently fail at runtime.

atymic (changes requested):

src/EpicGames/Provider.php
SGzuis: Fixed, thanks!

src/EpicGames/Provider.php
SGzuis: Fixed, thanks!

src/EpicGames/README.md
SGzuis: Fixed, thanks!

atymic (approved):

PR #1473, opened 2026-08-15
feat: add Frappe provider

kayedspace:
## Summary

  Adding Laravel Socialite provider for Frappe-based applications, including ERPNext, Frappe HR, LMS, Helpdesk, Books, and custom Frappe apps.

  ## Changes

  - Add OAuth 2.0 / OpenID Connect authorization
  - Support configurable Frappe instances through `base_url`
  - Exchange authorization codes for access and refresh tokens
  - Fetch and map the authenticated user profile
  - Support `userFromToken()` and token refresh
  - Use `openid` as the default scope
  - Support custom User fields through the `fields` configuration
  - Automatically request the `all` scope when custom fields are configured
  - Add OAuth token revocation through `revokeToken()`
  - Add the Frappe browser logout URL through `getLogoutUrl()`
  - Document installation, configuration, custom fields, logout, and token revocation
  - Add provider tests and user fixtures

  ## Testing

  - OAuth login tested against a live Frappe / ERPNext instance
  - User profile and custom field retrieval tested
  - Access-token user fetching tested
  - Refresh-token flow tested
  - Token revocation and browser logout endpoint tested
  - Full test suite passes
  - Laravel Pint passes

### DIFF
--- a/src/Frappe/Provider.php
+++ b/src/Frappe/Provider.php
@@ -1,0 +1,145 @@
+<?php
+
+namespace SocialiteProviders\Frappe;
+
+use GuzzleHttp\RequestOptions;
+use Illuminate\Support\Arr;
+use InvalidArgumentException;
+use Psr\Http\Message\ResponseInterface;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    public const string IDENTIFIER = 'FRAPPE';
+
+    protected $scopeSeparator = ' ';
+
+    protected $scopes = ['openid'];
+
+    public static function additionalConfigKeys(): array
+    {
+        return ['base_url', 'fields'];
+    }
+
+    /** {@inheritDoc} */
+    public function getScopes()
+    {
+        return array_values(array_unique(array_merge(
+            $this->scopes,
+            $this->getConfig('fields', []) ? ['all'] : []
+        )));
+    }
+
+    public function getLogoutUrl(): string
+    {
+        return $this->getBaseUrl().'/api/method/logout';
+    }
+
+    public function revokeToken(string $token): ResponseInterface
+    {
+        return $this->getHttpClient()->post(
+            $this->getBaseUrl().'/api/method/frappe.integrations.oauth2.revoke_token',
+            [RequestOptions::FORM_PARAMS => ['token' => $token]]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function refreshToken($refreshToken)
+    {
+        $response = $this->getHttpClient()->post($this->getTokenUrl(), [
+            RequestOptions::FORM_PARAMS => [
+                'grant_type'    => 'refresh_token',
+                'refresh_token' => $refreshToken,
+                'client_id'     => $this->clientId,
+                'client_secret' => $this->clientSecret,
+            ],
+        ]);
+
+        return json_decode((string) $response->getBody(), true);
+    }
+
+    protected function getTokenUrl(): string
+    {
+        return $this->getBaseUrl().'/api/method/frappe.integrations.oauth2.get_token';
+    }
+
+    protected function getAuthUrl($state): string
+    {
+        return $this->buildAuthUrlFromBase(
+            $this->getBaseUrl().'/api/method/frappe.integrations.oauth2.authorize',
+            $state
+        );
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


src/Frappe/Provider.php
```diff
@@ -0,0 +1,145 @@
...
+use GuzzleHttp\RequestOptions;
+use Illuminate\Support\Arr;
+use InvalidArgumentException;
+use Psr\Http\Message\ResponseInterface;
+use SocialiteProviders\Manager\OAuth2\AbstractProvider;
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    public const string IDENTIFIER = 'FRAPPE';
```
atymic, line 14: `public const string` is 8.3+ but composer says `^8.0`, so this fatals on 8.0-8.2. rather than dropping the type, can you bump the php constraint to `^8.3`? we only support supported php vers anyway. suggestion is on composer.json.

src/Frappe/composer.json
```diff
@@ -0,0 +1,34 @@
...
+            "email": "3likayed@gmail.com"
+        }
+    ],
+    "support": {
+        "issues": "https://github.com/socialiteproviders/providers/issues",
+        "source": "https://github.com/socialiteproviders/providers",
+        "docs": "https://socialiteproviders.com/frappe"
+    },
+    "require": {
+        "php": "^8.0",
```
atymic: 8.3+ please, we only support supported php vers (and the typed const needs it)

```suggestion
        "php": "^8.3",
```

src/Frappe/README.md
```diff
@@ -0,0 +1,145 @@
...
+Revoke the OAuth access token and log out of your Laravel app:
+
+```php
+Socialite::driver('frappe')->revokeToken($user->token);
+Auth::logout();
+```
+
+To also end the user's Frappe browser session, submit a `POST` form to
+`Socialite::driver('frappe')->getLogoutUrl()`. Do not redirect to this URL;
+current Frappe versions require `POST`. See Frappe's official
```
atymic: the docs you linked below say `GET /api/method/logout`, not POST. is the POST thing from a newer version? if so can you link that, otherwise this should probably say GET.

src/Frappe/Provider.php
```diff
@@ -0,0 +1,145 @@
...
+                'Authorization' => 'Bearer '.$token,
+            ],
+            RequestOptions::QUERY => [
+                'fields'            => json_encode($fields),
+                'filters'           => json_encode([['name', '=', Arr::get($user, 'email')]]),
+                'limit_page_length' => 1,
+            ],
+        ]);
+
+        $fields = json_decode((string) $response->getBody(), true)['data'][0] ?? [];
```
atymic: reusing `$fields` for the response here reads a bit odd since it was the config a few lines up. maybe `$extra`?

src/Frappe/Provider.php
```diff
@@ -0,0 +1,145 @@
...
+    public function revokeToken(string $token): ResponseInterface
+    {
+        return $this->getHttpClient()->post(
+            $this->getBaseUrl().'/api/method/frappe.integrations.oauth2.revoke_token',
+            [RequestOptions::FORM_PARAMS => ['token' => $token]]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
```
atymic: these `{@inheritdoc}` blocks aren't needed, can you drop them? (same on 92 and 130)

atymic (changes requested):

src/Frappe/Provider.php
kayedspace: Done, bumped to ^8.3

src/Frappe/Provider.php
kayedspace: Dropped them.

src/Frappe/Provider.php
kayedspace: Good catch.
renamed the variable to `$extra`

src/Frappe/README.md
kayedspace: you're right that it's a version thing.
 `/api/method/logout` was `@frappe.whitelist(allow_guest=True)` through v15, so GET works there.

 It was restricted to `methods=["POST"]` in frappe/frappe@9c6594b47c04fd17095dd9058d2b2792dce8de2 which on `develop/version-16`
but not `version-15` the docs page hasn't caught up yet.

I've rewritten the section to cover both versions and link the commit.

atymic (approved):

PR #1480, opened 2026-08-24
Allow configuring azure OAuth base URL

wheesnoza:
## Summary
This PR allows the Azure provider's OAuth base URL to be configured via `base_url`.

This follows the same use case as the recently added #1479 `graph_url` configuration: callers can point Azure requests at mock endpoints in local or test environments without subclassing the provider.

By default, the provider continues to use `https://login.microsoftonline.com`, so existing behavior remains unchanged.

## Changes

- Add `base_url` to Azure additional config keys
- Use configured `base_url` when building Azure OAuth URLs
- Document the optional config value
- Add Azure provider test coverage

### DIFF
--- a/src/Azure/Provider.php
+++ b/src/Azure/Provider.php
@@ -10,6 +10,13 @@ class Provider extends AbstractProvider
     public const IDENTIFIER = 'AZURE';
 
     /**
+     * The base Azure OAuth URL.
+     *
+     * @var string
+     */
+    protected $baseUrl = 'https://login.microsoftonline.com';
+
+    /**
      * The base Azure Graph URL.
      *
      * @var string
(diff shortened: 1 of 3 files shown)


atymic (changes requested):
Can you let me know what case you would do this?

wheesnoza:
@atymic 
The main use case is similar to the `graph_url` option added in #1479: local/test environments where an application wants to run the Azure OAuth flow against a trusted mock OAuth/OIDC server instead of the real Microsoft login endpoint.

In my company's project, we run a local mock of Entra ID so we can develop and test login without a real tenant. With `base_url` configurable, we can point Socialite at that mock through config alone, instead of subclassing the provider to override `getBaseUrl()`.

However, if you think this is too broad for the provider, I’m happy to adjust or close it.

PR #1485, opened 2026-09-05
Validate Apple callback state

binuka200:
## Summary
- Validate OAuth state before exchanging the Apple authorization code
- Remove reconstruction of the expected state from the callback token
- Add callback state regression coverage

## Testing
- `vendor/bin/phpunit tests/Apple/AppleProviderTest.php --testdox`

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -276,26 +276,15 @@ class Provider extends AbstractProvider
      */
     public function user()
     {
-        //Temporary fix to enable stateless
+        if ($this->usesState() && $this->hasInvalidState()) {
+            throw new InvalidStateException;
+        }
+
         $response = $this->getAccessTokenResponse($this->getCode());
 
         $appleUserToken = $this->getUserByToken(
             $token = Arr::get($response, 'id_token')
         );
-
-        if ($this->usesState()) {
-            $state = explode('.', $appleUserToken['nonce'])[1];
-            if ($state === $this->request->input('state')) {
-                $this->request->session()->put([
-                    'state'        => $state,
-                    'state_verify' => $state,
-                ]);
-            }
-
-            if ($this->hasInvalidState()) {
-                throw new InvalidStateException;
-            }
-        }
 
         $user = $this->mapUserToObject($appleUserToken);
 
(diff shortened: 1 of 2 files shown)


src/Apple/Provider.php
```diff
@@ -276,27 +276,16 @@ public static function verify($jwt)
      */
     public function user()
     {
-        //Temporary fix to enable stateless
+        if ($this->usesState() && $this->hasInvalidState()) {
+            throw new InvalidStateException;
```
atymic: this is the right shape, and moving it before `getAccessTokenResponse()` is a nice side effect, a forged callback no longer burns a token exchange against Apple.

worth knowing `hasInvalidState()` already returns false when `isStateless()`, so the `usesState() &&` is belt and braces. no harm in it, just noting it's not load bearing.

tests/Apple/CallbackStateValidationTest.php
```diff
@@ -0,0 +1,43 @@
...
+    public function test_callback_without_previously_stored_state_is_rejected(): void
+    {
+        $request = $this->makeRequestWithSession([
+            'code'  => 'authorization-code',
+            'state' => 'attacker-controlled-state',
+        ]);
+
+        $this->expectException(InvalidStateException::class);
+
+        $this->makeAppleProvider($request)->user();
```
atymic: this one passes without a mocked http client, so it'd still go green if the state check were gone (it'd just blow up on the guzzle call instead). can you make it fail for the state reason only, queueing a response does it:

```suggestion
        $provider = $this->makeAppleProvider($request);
        $provider->setHttpClient($this->makeHttpClient([
            new Response(200, [], json_encode([
                'id_token' => $this->identityToken(['nonce' => 'nonce']),
            ])),
        ]));

        $this->expectException(InvalidStateException::class);

        $provider->user();
```

tests/Apple/CallbackStateValidationTest.php
```diff
@@ -0,0 +1,43 @@
...
+                'id_token' => $this->identityToken(['nonce' => 'nonce']),
+            ])),
+        ]));
+
+        $user = $provider->user();
+
+        $this->assertSame('apple-user-id', $user->getId());
+        $this->assertSame('user@example.com', $user->getEmail());
+    }
+}
```
atymic: could you add one for the cross site post case? empty session but a state that would have matched, ie the real user whose cookie got dropped by samesite=lax. that's the behaviour change people will actually hit, and having it pinned in a test makes the README note obvious to whoever reads it next.

atymic (approved):

PR #1486, opened 2026-09-07
feat(vipps): add Vipps OAuth2 provider

Johannes-Andersen:


src/Vipps/Provider.php
```diff
@@ -0,0 +1,148 @@
...
+            return $configuration;
+        });
+    }
+
+    protected function getHeaders(): array
+    {
+        $headers = [
+            'Accept'                       => 'application/json',
+            'Vipps-System-Name'            => 'laravel-socialite',
+            'Vipps-System-Version'         => InstalledVersions::getPrettyVersion('laravel/socialite'),
```
atymic: you guard the plugin version with `isInstalled` on the line below but not this one. `getPrettyVersion` throws `OutOfBoundsException` if the package isn't there, and it can return null even when it is (root package / replace), which guzzle won't accept as a header value.

can we do the same treatment here?

```suggestion
            'Vipps-System-Version'         => InstalledVersions::isInstalled('laravel/socialite')
                ? (InstalledVersions::getPrettyVersion('laravel/socialite') ?? 'dev')
                : 'dev',
```

src/Vipps/Provider.php
```diff
@@ -0,0 +1,148 @@
...
+            'Vipps-System-Plugin-Name'     => 'socialiteproviders-vipps',
+            'Vipps-System-Plugin-Version'  => InstalledVersions::isInstalled('socialiteproviders/vipps')
+                ? InstalledVersions::getPrettyVersion('socialiteproviders/vipps')
```
atymic: same here, `getPrettyVersion` can return null so this can still end up as a null header. `?? 'dev'` would cover it.

src/Vipps/composer.json
```diff
@@ -0,0 +1,33 @@
...
+        }
+    ],
+    "support": {
+        "issues": "https://github.com/socialiteproviders/providers/issues",
+        "source": "https://github.com/socialiteproviders/providers",
+        "docs": "https://socialiteproviders.com/vipps"
+    },
+    "require": {
+        "php": "^8.3",
+        "socialiteproviders/manager": "^4.4"
```
atymic: you're using the `Cache` facade in the provider, so can we add illuminate/support here please?

```suggestion
        "illuminate/support": "^11.0 | ^12.0 | ^13.0",
        "socialiteproviders/manager": "^4.4"
```

src/Vipps/Provider.php
```diff
@@ -0,0 +1,148 @@
...
+        if (! is_string($user['sub'] ?? null) || $user['sub'] === '') {
+            throw new UnexpectedValueException('Vipps Userinfo response is missing sub.');
+        }
+
+        return (new User)->setRaw($user)->map([
+            'id'       => $user['sub'],
+            'nickname' => null,
+            'name'     => $user['name'] ?? null,
+            'email'    => $user['email'] ?? null,
+            'avatar'   => null,
```
atymic: vipps doesn't return an avatar or a nickname, can we just drop these two keys rather than hardcoding null?

src/Vipps/Provider.php
```diff
@@ -0,0 +1,148 @@
...
+        ];
+
+        if ($msn = $this->getConfig('merchant_serial_number')) {
+            $headers['Merchant-Serial-Number'] = $msn;
+        }
+
+        return $headers;
+    }
+
+    protected function usesBasicAuthentication(): bool
```
atymic: minor, but `usesBasicAuthentication` reads like a socialite hook and isn't one. `tokenAuthMethod()` or similar might be clearer. also this validates config on every call, is it worth throwing once rather than from three call sites?

src/Vipps/Provider.php
```diff
@@ -0,0 +1,148 @@
...
+        return array_values(array_unique(array_merge(['openid'], parent::getScopes())));
+    }
+
+    protected function getOpenIdConfiguration(): array
+    {
+        $baseUrl = filter_var($this->getConfig('test_mode', false), FILTER_VALIDATE_BOOLEAN)
+            ? 'https://apitest.vipps.no'
+            : 'https://api.vipps.no';
+
+        return Cache::remember('socialiteproviders.vipps.openid.'.md5($baseUrl), 3600, function () use ($baseUrl) {
```
atymic: the base url is already in the key via the md5, is there a reason not to just use `test`/`prod` in the key? would make it a lot easier to bust manually if the shape ever changes.

src/Vipps/Provider.php
atymic: `getPrettyVersion` throws `OutOfBoundsException` if the package isn't installed, and it can return null even when it is (root package / replace), which guzzle won't accept as a header value. the plugin one below is guarded for the first case but can still resolve to null.

```suggestion
            'Vipps-System-Version'         => InstalledVersions::isInstalled('laravel/socialite')
                ? (InstalledVersions::getPrettyVersion('laravel/socialite') ?? 'dev')
                : 'dev',
            'Vipps-System-Plugin-Name'     => 'socialiteproviders-vipps',
            'Vipps-System-Plugin-Version'  => InstalledVersions::isInstalled('socialiteproviders/vipps')
                ? (InstalledVersions::getPrettyVersion('socialiteproviders/vipps') ?? 'dev')
                : 'dev',
```

src/Vipps/composer.json
atymic, line 27: you're using the `Cache` facade in the provider, so can we add illuminate/support here please?

```suggestion
        "illuminate/support": "^11.0 | ^12.0 | ^13.0",
        "socialiteproviders/manager": "^4.4"
```

src/Vipps/Provider.php
atymic: vipps doesn't return an avatar or a nickname, can we just drop these two keys rather than hardcoding null?

atymic (commented):
Thanks for this, really nicely done. Tests, discovery, pint and composer validate all pass here, and the docs link made it easy to check the endpoints against.

Two things before it goes in: the version headers can throw or send null, and composer needs illuminate/support since you're using the Cache facade.

src/Vipps/Provider.php
Johannes-Andersen: Sure! Noticed the other where defaulting to null, so I just did the same to avoid any potential issues with leaving it blank. But happy to just remove :D 

Screenshot 2026-09-09 at 15 14 11
Screenshot 2026-09-09 at 15 14 31

PR #1488, opened 2026-09-10
fix(kanidm): enable PKCE using intended mechanism

haydenmc:
A previous change (#1421) enabled PKCE for Kanidm by turning it on in the `getBaseUrl()` function.

This can cause authentication to fail if `usesPKCE()` is called *before* `getBaseUrl()`, which is the case when using Kanidm with [Lychee](https://github.com/LycheeOrg/Lychee), resulting in a 500 error during the auth flow, and this error in Kanidm logs:

```
kanidm-1       | 6db89809-5bcc-4469-bd6a-54d717ec98d1 INFO     ┕━ handle_oauth2_token_exchange [ 123µs | 54.21% ]
kanidm-1       | 6db89809-5bcc-4469-bd6a-54d717ec98d1 INFO        ┕━ ｉ [info]: PKCE code verification failed - code challenge is present, but no verifier was provided | event_tag_id: 10
```

AbstractProvider.php is meant to populate `code_verifier` [here](https://github.com/laravel/socialite/blob/fa0181ee6204ca28a55cd67145fadd631ac209cf/src/Two/AbstractProvider.php#L169-L171) but doesn't, due to `usesPKCE()` returning false.

This change adds the explicit function override for `usesPKCE()`, which will properly indicate to socialite that PKCE is enabled.

### DIFF
--- a/src/Kanidm/Provider.php
+++ b/src/Kanidm/Provider.php
@@ -34,15 +34,16 @@ class Provider extends AbstractProvider
             throw new InvalidArgumentException('Missing base URL value.');
         }
 
-        /**
-         * PKCE is the default in recent versions of Kanidm, and heavily encouraged
-         * to be enforced whenever possible.
-         */
-        if (!$this->usesPKCE() && $this->getConfig('enable_pkce')) {
-            $this->enablePKCE();
-        }
-
         return $baseUrl;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


atymic (approved):
Could you add a test please?

haydenmc:
Added tests inspired by some from another provider (Vipps) and confirmed they are passing locally.

Please let me know if you need anything else. Thank you!

PR #1489, opened 2026-09-10
fix(apple): verify the callback nonce

atymic:
Follow up to #1485. That PR fixed the state check but left the nonce unverified on the web callback: `getCodeFields()` still sent one, and `user()` never looked at it, so any Apple-signed identity token for our client id was accepted once the state matched. #1484 added nonce verification for the native flow only.

- issue a random nonce on `redirect()` and keep it in the session next to the state
- pull it on the callback and pass it to `checkToken()`, so a token from another authorization request is rejected
- drop the `usesState()` guard around `hasInvalidState()`, it already returns false when stateless
- document the `SameSite` requirement in the README

The nonce is only checked when the session holds one, so logins in flight during the deploy still complete.

**Upgrade note**

Apple posts the callback cross-site. Laravel's default `SESSION_SAME_SITE=lax` cookie isn't sent on that POST, so the session is empty and the state check from #1485 rejects every login. Before #1485 the provider rebuilt the state from the token and let it through, which is the CSRF hole. Users need `SESSION_SAME_SITE=none` with `SESSION_SECURE_COOKIE=true`. That's the breaking change, so this should ship as `release:major` along with #1485 (which merged unlabelled and hasn't released yet).

**Testing**

- `vendor/bin/phpunit tests/Apple` (23 tests)
- full suite passes, 271 tests
- pint failures on `src/Apple/Provider.php` are pre-existing on master

### DIFF
--- a/src/Apple/Provider.php
+++ b/src/Apple/Provider.php
@@ -72,6 +72,18 @@ class Provider extends AbstractProvider
     /**
      * {@inheritdoc}
      */
+    public function redirect()
+    {
+        if ($this->usesState()) {
+            $this->request->session()->put('nonce', Str::random(40));
+        }
+
+        return parent::redirect();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function getCodeFields($state = null)
     {
         $fields = [
@@ -84,7 +96,7 @@ class Provider extends AbstractProvider
 
         if ($this->usesState()) {
             $fields['state'] = $state;
-            $fields['nonce'] = Str::uuid() . '.' . $state;
+            $fields['nonce'] = $this->request->session()->get('nonce');
         }
 
         return array_merge($fields, $this->parameters);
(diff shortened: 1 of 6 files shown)


atymic:
@binuka200 could you have a look at this please?

binuka200:
Thanks for asking me to take a look. I’ve checked the latest changes against Apple’s documentation and RFC 9700.

The normal session-based flow looks good to me. Generating and storing a fresh nonce, then rejecting the callback when it is missing or doesn’t match, should close the remaining gap.

I’m less sure about the new statelessNonce() approach. Because the nonce is kept in a shared cache using the returned state as the key, it doesn’t seem to be tied to the browser that originally started the login.

For example, an attacker could start an Apple login themselves, capture their complete callback, and then submit it through another user’s browser. The cached nonce would still match, even though that browser didn’t start the login.

RFC 9700 says the nonce needs to be securely bound to the user agent where the login started:

https://www.rfc-editor.org/rfc/rfc9700.html#section-2.1.1

So I think the session-based part is ready, but the stateless cache flow may need another browser-bound value, such as a short-lived Secure, HttpOnly, SameSite=None cookie. What do you think? @atymic

atymic:
@binuka200 have swapped it over to a http only cookie, let me know what you think

binuka200:
Thanks for making the changes. This looks good to me now.

I use Sign in with Apple in PHP quite regularly, so I’ll keep an eye on the provider and contribute improvements where I can.

